PR #12, opened 2015-02-04
Clarify that refunds does not "free up" capturable amount

ct-clearhaus:
Clear enough to eliminate the misunderstanding QP seems to have gotten?
@x2q @soltveit


ct-clearhaus:
Poke @soltveit

ct-clearhaus:
Merging as this seems to be a "normal" corner case which implementers should be aware of; otherwise we'll supposedly just get support cases asking for this kind of clarification.

PR #16, opened 2015-02-25
Other explanations for idempotency

ct-clearhaus:
I'm not sure what is better.

Judging what's easier to understand vs. what's more precise is a tough one.


ct-clearhaus:
IIRC it was done.

PR #18, opened 2015-03-26
Notice about acquirer_bin and merchant_id

ct-clearhaus:
Before this is merged, we should have an endpoint where a merchant can `GET` the merchant ID. And, then the text should be adjusted according to that.

### DIFF
--- a/source/index.md
+++ b/source/index.md
@@ -340,6 +340,20 @@ Example response (snippet):
 }
 ````
 
+<p class="alert alert-info">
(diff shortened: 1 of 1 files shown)


ct-clearhaus:
We actually have multiple bins, and possibly we need to spread the merchants across these depending on :grey_question: ... So I'll kill this. We'll put it all in the endpoint where a merchant can get the merchant ID.

PR #21, opened 2015-05-18
Add HMAC

ct-clearhaus:
RFR @soltveit @jsmpereira

### DIFF
--- a/source/index.md
+++ b/source/index.md
@@ -81,7 +81,8 @@ Number  Text
 
 Requests can optionally be signed.
 
-The signature is an RSA or HMAC signature represented in Hex of the HTTP body.
+The signature is an RSA signature or HMAC of the HTTP body; it is represented in
+Hex.
 The signee must be identified by the signing API-key. Both should be provided in
 a `Signature` header together with either `RS256-hex` or `HS256-hex` identifying
 the signing algorithm:
--- a/source/index.md
+++ b/source/index.md
@@ -81,14 +81,48 @@ Number  Text
 
 Requests can optionally be signed.
 
-The signature is an RSASSA-PKCS1-v1_5 signature of the body represented in Hex.
-The signee must be identified by the signing API-key. Both should be provided
-in a `Signature` header together with `RS256-hex` identifying the signing
-algorithm:
+The signature is an RSA or HMAC signature represented in Hex of the HTTP body.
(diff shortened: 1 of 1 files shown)


source/index.md
```diff
@@ -81,15 +81,49 @@ Number  Text
 
 Requests can optionally be signed.
 
-The signature is an RSASSA-PKCS1-v1_5 signature of the body represented in Hex.
-The signee must be identified by the signing API-key. Both should be provided
-in a `Signature` header together with `RS256-hex` identifying the signing
-algorithm:
+The signature is an RSA or HMAC signature represented in Hex of the HTTP body.
```
jsmpereira: Reads a bit weird to me.
How about like before: "The signature is an RSA or HMAC signature of the HTTP body represented in Hex."

jsmpereira:
Is it relevant to mention somewhere that singing a request allows you to **_not_** send the CSC?

source/index.md
ct-clearhaus: I agree. But it's possible to read like "a signature of something; that something is the HTTP body represented in Hex". How one then would represent the HTTP body in Hex ... I dunno.

ct-clearhaus:
> Is it relevant to mention somewhere that singing a request allows you to not send the CSC?

Nah. Because it may also allow you to let others use your card token. It's not a "all or nothing" neither wrt. CSC nor wrt. cross card usage, so I think we better leave it minimal and let those that care know.

source/index.md
jsmpereira: I agree as well :stuck_out_tongue:

ct-clearhaus:
That's better IMHO. Agree @jsmpereira?

jsmpereira:
:+1:

source/index.md
```diff
@@ -81,15 +81,50 @@ Number  Text
...
+
+```
+OpenSSL::HMAC.hexdigest('sha256', hmac_secret, body)
+```
+
+
+### RSA signature
+
+The RSA signature is an RSASSA-PKCS1-v1_5 signature of the body. It is
+represented in Hex.
```
jsmpereira: We had the answer right here :stuck_out_tongue:

ct-clearhaus:
@soltveit Will you deploy soon?

PR #22, opened 2015-09-18
Three very small fixes.

mt-clearhaus:
This should probably just be merged in without hesitation :-)

@ct-clearhaus It seems you authored most recently. Ok? :-)


ct-clearhaus:
:+1:

PR #31, opened 2016-08-02
Explain how to trigger different error messages

ct-clearhaus:
This is _not_ supported by the test gateway as for now! But wouldn't this be nice? Is it a feature we want to add?

Depends on https://github.com/clearhaus/gateway-api-docs/pull/30 (or something alike).


mt-clearhaus:
I like it. It will require additional details and some accompanying code, though :smile:

ct-clearhaus:
Too old.

PR #34, opened 2016-10-06
Adjust card representation

ct-clearhaus:
* scheme may also be `unknown`
* country is omittable
* type is omittable


ct-clearhaus:
RFR @soltveit

soltveit:
👍

PR #36, opened 2016-11-08
Remove bin from example card response

ct-clearhaus:
RFR @mt-clearhaus @soltveit

### DIFF
--- a/source/index.md
+++ b/source/index.md
@@ -643,8 +643,6 @@ authorization is made when POSTing to this endpoint.</p>
 #### Response parameters
 
 <dl class="dl-horizontal">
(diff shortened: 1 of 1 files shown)


mt-clearhaus:
@ct-clearhaus tells me we have never responded with `bin` :smile: 

:+1:

ct-clearhaus:
Obviously nobody is missing it, so I think we should update the docs as this PR suggests rather than fix the code.

jsmpereira:
@ct-clearhaus 
I suppose `bin` should also be removed from `Response parameters` at http://docs.gateway.clearhaus.com/#cards?

PR #41, opened 2017-04-25
Restrict text_on_statement

ct-clearhaus:


ct-clearhaus:
> If we reject based on a rule involving text on statement will we also reject with status code "Invalid text on statement"?

Nope. Then it's a rule violation. The invalid text on statement is one that doesn't follow the API spec. This is common for all merchants whereas the rules differ between merchants.

soltveit:
Right, makes sense.

source/index.md
```diff
@@ -540,6 +542,7 @@ POST https://gateway.clearhaus.com/authorizations/:id/refunds
   <dd>
     [\x20-\x7E]{0,22}
     <i><a target="_blank" href="http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">ASCII printable characters</a></i> </br>
+    <i>May not be all digits, all same character, or all sequential characters (e.g. "abc").</i></br>
     <i>Optional</i> <br />
     Text that will be placed on cardholder's bank statement.
```
mt-clearhaus: Maybe add "Overrides <code>text_on_statement</code> from authorization."?

ct-clearhaus: Please feel free to push 🙂

ct-clearhaus:
Last push due to https://github.com/clearhaus/gateway-api-docs/pull/43#issuecomment-296962040

ct-clearhaus:
How much time does PSPs want in order to handle an update like this to add a response status code?

I've checked the last 30 days of transactions finding only very few transactions that would be hit by this; I've informed you who've sent us those transactions.

Cc: @nielsek @ta @tjconcept @storkie

ta:
@ct-clearhaus: around 0 :)

tjconcept:
It's gracefully handled already. No immediate changes required here.

mt-clearhaus:
Deployed in production today in the timespan 15:15-16:00 CEST.

PR #42, opened 2017-04-25
Account descriptor as `descriptor`

ct-clearhaus:
This is for consistency. Actually, `GET /account` is the only place we screwed up and named the account's descriptor `text_on_statement`.

The naming is that the account has a `descriptor` whereas a transaction has a `text_on_statement`. The account's `descriptor` is the default `text_on_statement` for all transactions if nothing is supplied in the `POST` to create the given transaction.

### DIFF
--- a/source/index.md
+++ b/source/index.md
@@ -411,7 +411,7 @@ Example response (snippet):
 {
     "merchant_id": "000000003000001",
     "name": "Merchant Ltd.",
-    "text_on_statement": "merchant.com",
+    "descriptor": "merchant.com",
     "country": "GBR",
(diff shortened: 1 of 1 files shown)


ct-clearhaus:
I know some gateways do fetch it. We poke them to deprecate it after we've added `descriptor`, give them a few weeks to adjust, and then remove `text_on_statement`.

mt-clearhaus:
The addition of `descriptor` has been deployed in production today in the timespan 15:15-16:00 CEST. We are now returning both `descriptor` and `text_on_statement`.

PR #44, opened 2017-04-27
Require expire_year in this century

ct-clearhaus:
We're seeing `19\d\d` be input as expire year and other "funny" numbers that are accepted as cards that will obviously never result in a successful transaction. Thus, let us throw these cards away earliest possible.


mt-clearhaus:
Deployed in production today in the timespan 15:15-16:00 CEST.

PR #45, opened 2017-06-08
Rough draft of "a subscription model" to get the ball rolling.

ad-clearhaus:
This is very much a rough draft just to have something to work on.
The more I added to the description, the more unanswered questions and incoherence popped up. 

So lets talk about that. :)

### DIFF
--- a/source/index.md
+++ b/source/index.md
@@ -315,30 +315,76 @@ Example response (snippet):
 ````
 
 
-## Recurring payments
+##  Subscription concept
 
-Recurring payments enable you to repeatedly charge cardholders without having
+Subscription payments enable you to repeatedly charge cardholders without having
 them to provide card information for subsequent payments.
 
-### Subscription concept
+Subscription paymests also has an `update` capability, that allows the subsequent
+paymenst to continue eventhough the cardholder changes card information.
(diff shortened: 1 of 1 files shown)


ad-clearhaus:
In the draft model, a subscription share many similarities with regular auth/capt. The main difference is that the url's become /subscriptions/<uuid-id>/ longer.
Describing how the authorize and capture a subscription seemed redundant, and will contain pitfalls. The two methods are very similar and the subtle differences can be hard for us to write short and concise, and easy for implementers to miss.

ad-clearhaus:
At the moment I am leaning towards a model where a `subscription` will create a uniq `card`-object that is scoped only to work on that particular `subscription`. Subsequent transaction on that subscription would follow the existing pattern of recurring transaction (e.g. `POST /cards/:id/auth...`), and append `ss_id` to the parameters. Creating the subscription returns the uniq `cardid`. Calling `/subscriptions/:id/update` would, in case of VAU update, create a new card and return it in `/subscription/:id/update/response`. `GET /subscription/:id/` could also return the current `cardid` on the subscription.

ad-clearhaus:
For internal notes on VAU/ABU see https://github.com/clearhaus/updatr/wiki

PR #53, opened 2017-11-02
Payment methods and other updates

ct-clearhaus:
* Add payment methods: `card`, `applepay`, `mobilepayonline`
* Adjust description of recurring payments
* Versioning
* Add `transaction_rules` and `acquirer` to example response to `GET /account`
* Deprecate `text_on_statement` in the response to `GET /account`
* Rename of `card[number]` to `card[pan]`
* Deprecate `card[number]` except on `/cards` which is entirely being
  deprecated
* Describe these changes in the Changes section


source/index.md
```diff
@@ -763,10 +855,13 @@ Declined   |  40000 |  General input error
 
 ### Status message
 
-A status message is included in every response when you make a new transaction.
+A status message may be included in every response when you create a new
```
jsmpereira: The use of `may` together with `every` sounds weird here. How about `s/every/the`?

source/index.md
```diff
@@ -728,6 +814,12 @@ https://gateway.clearhaus.com/account
   <dd>[0-9]{4} <br /> Merchant category code.</dd>
   <dt>acquirer</dt>
   <dd>Used for 3-D Secure.</dd>
+  <dt>transaction_rules</dt>
+  <dd>[\x20-\x7E]* <br /> The transaction processing rules that the merchant's transactions must adhere to.</dd>
```
jsmpereira: Maybe `s/The transaction processing rules/The processing rules`.

source/index.md
```diff
@@ -492,22 +521,80 @@ POST https://gateway.clearhaus.com/authorizations
...
 </dl>
 
 <p class="alert alert-info">
-<b>Notice:</b> The card dictionary is not required when making an authorization
-based on a card resource. However, <code>card[csc]</code> should be populated
-when CSC is available.
+  <b>Notice:</b> An authorization that include <code>card[pares]</code> is 3-D
+  Secured and is considered multi-factor authenticated.
+  <br />
+  <b>Notice:</b> An authorization that include <code>card[pares]</code> and/or
```
jsmpereira: `s/that include/that includes`.

source/index.md
```diff
@@ -492,22 +521,80 @@ POST https://gateway.clearhaus.com/authorizations
...
     A reference to an external object, such as an order number.
   </dd>
+
+  <!-- deprecated -->
   <dt>threed_secure[pares]</dt>
-  <dd>[:base64:] <br /> <i>Optional</i> <br /> See more information on <a target="_blank" href="http://docs.3dsecure.io">3Dsecure.io</a></dd>
+  <dd>
+    Deprecated! Please use <code>card[pares]</code>. <br />
+    [:base64:] <br />
+    See more information on <a target="_blank" href="http://docs.3dsecure.io">3Dsecure.io</a>.
```
jsmpereira: `s/more information on/more information at` for consistency. Ref. https://github.com/clearhaus/gateway-api-docs/pull/53/files#diff-eeb8bdec9c0eef052c0532fce07c4bacR402.

source/index.md
```diff
@@ -492,22 +521,80 @@ POST https://gateway.clearhaus.com/authorizations
...
+  <dt>card[pan]</dt>
   <dd>[0-9]{12,19} <br /> Primary account number of card to charge.</dd>
   <dt>card[expire_month]</dt>
   <dd>[0-9]{2} <br /> Expiry month of card to charge.</dd>
   <dt>card[expire_year]</dt>
   <dd>20[0-9]{2} <br /> Expiry year of card to charge.</dd>
   <dt>card[csc]</dt>
   <dd>[0-9]{3} <br /> Card Security Code.</dd>
+  <dt>card[pares]</dt>
+  <dd>[:base64:] <br /> <i>Optional</i> <br /> See more information on <a target="_blank" href="http://docs.3dsecure.io">3Dsecure.io</a></dd>
```
jsmpereira: `s/more information on/more information at`

source/index.md
```diff
@@ -372,18 +380,29 @@ Example response (snippet):
...
    "processed_at": "2014-07-09T13:33:44+00:00",
-   "recurring": true
+   "recurring": true,
+   "_embedded": { "card": { "id": "58dabba0-e9ea-4133-8c38-bfa1028c1ed2" } }
 }
 ````
 
-Above can be repeated whenever you need to reserve money and must be followed
-by a capture transaction.
+This should (in case of non-zero amount) be followed by a capture.
```
jsmpereira: Another way to write this could be:

> In case of a non-zero amount this should be followed by a capture.

source/index.md
```diff
@@ -372,18 +380,29 @@ Example response (snippet):
...
+
+Subsequent authorizations are made similarly, but obviously cannot include
```
jsmpereira: Maybe `would not` works here as well, instead of `cannot`. In the sense that, subsequent _would not_ have the cardholder present.

jsmpereira: > Subsequent authorizations are made similarly, but obviously neither CSC nor PARes can/would be included. Use the [cards](#cards) endpoint or, when approved for CSC-less transactions, the `card` payment method.

source/index.md
ct-clearhaus: What about

> This should be followed by a capture except when the amount is `0`.

?

source/index.md
```diff
@@ -372,18 +380,29 @@ Example response (snippet):
...
+neither CSC nor a PARes; either using the [cards](#cards) endpoint or by (when
+approved for CSC-less transactions) using the `card` payment method.
+
+When using the `applepay` payment method, the first recurring transaction is
+made using `applepay[*]` whereas subsequent recurring transactions are made using
+`card[]` with the details from the first transaction's `applepay[token]`.
```
ct-clearhaus: Inconsistent to use "`applepay[*]`" and `card[]`.

Also, one could use the responded card token or, when approved for CSC-less transactions, the card details in the first transaction's `applepay[token]`.

source/index.md
```diff
@@ -372,18 +380,30 @@ Example response (snippet):
...
+   "recurring": true,
+   "_embedded": { "card": { "id": "58dabba0-e9ea-4133-8c38-bfa1028c1ed2" } }
 }
 ````
 
-Above can be repeated whenever you need to reserve money and must be followed
-by a capture transaction.
+This should be followed by a capture except when the amount is `0`.
+
+Subsequent authorizations are made similarly, but obviously neither CSC nor
```
ad-clearhaus: We should flip the order of **Recurring payments** and **3-D Secure** since **Recurring payments** reference `pares`.
A note "_See 3-D Secure for information on `pares`_" would do as well.


But writing "_obviously_ neither CSC nor PARes would be included" is a bit on the nose, since the reader have not read the **3-D Secure section** yet, and familiarity  of `PARes` is not as widescread as `PAN` or `CSC`, so including/excluding `PARes` is not obvious. :)

ct-clearhaus:
It is possible for a PSP to have a flow where the phone number never reaches the PSP. The phone number can be "optional" in the same sense that CSC can be optional for trusted partners.

source/index.md
```diff
@@ -849,6 +945,17 @@ https://gateway.clearhaus.com/account
 
 ## Changes
 
+### Add payment methods and subscriptions
```
ct-clearhaus: Subscriptions is not added by this PR.

source/index.md
```diff
@@ -492,22 +522,80 @@ POST https://gateway.clearhaus.com/authorizations
...
+  <dt>card[pares]</dt>
+  <dd>[:base64:] <br /> <i>Optional</i> <br /> See more information at <a target="_blank" href="http://docs.3dsecure.io">3Dsecure.io</a></dd>
 </dl>
 
 <p class="alert alert-info">
-<b>Notice:</b> The card dictionary is not required when making an authorization
-based on a card resource. However, <code>card[csc]</code> should be populated
-when CSC is available.
+  <b>Notice:</b> An authorization that includes <code>card[pares]</code> is 3-D
+  Secured and is considered multi-factor authenticated.
```
mt-clearhaus: Maybe phrase this a bit differently in order to account for unauthenticated, attempt and error PARes'?

source/index.md
```diff
@@ -492,22 +522,80 @@ POST https://gateway.clearhaus.com/authorizations
...
+
+<dl class="dl-horizontal">
+  <dt>mobilepayonline[pan]</dt>
+  <dd>[0-9]{12,19} <br /> Primary account number of card to charge.</dd>
+  <dt>mobilepayonline[expire_month]</dt>
+  <dd>[0-9]{2} <br /> Expiry month of card to charge.</dd>
+  <dt>mobilepayonline[expire_year]</dt>
+  <dd>[0-9]{4} <br /> Expiry year of card to charge.</dd>
+  <dt>mobilepayonline[phonenumber]</dt>
+  <dd>[\x20-\x7E]{1,15} <br /> Phone number from where the PAN originates.</dd>
```
mt-clearhaus: @ct-clearhaus tells me this should be optional.

source/index.md
```diff
@@ -849,6 +945,17 @@ https://gateway.clearhaus.com/account
 
 ## Changes
 
+### Add payment methods and subscriptions
+
+Payment methods has been added to the authorization resource. Also, a few
```
mt-clearhaus: `s/has been/have been/`

source/index.md
```diff
@@ -849,6 +945,17 @@ https://gateway.clearhaus.com/account
...
+parameters and an endpoint has been deprecated. Refer to [the documentation
```
mt-clearhaus: `s/has been/have been/`

source/index.md
ct-clearhaus: "An *approved* authorization that ..."?

mt-clearhaus: :+1:

ct-clearhaus:
It's unclear to me if a subsequent recurring transaction should also have the DWO identifier set, and if a non-recurring, credential on file, merchant initiated transaction should also have the DWO identifier set. The latter would be problematic wrt. `POST /cards` — because the same card can arrive both via MPO and "normally".
We could require `POST /cards` to have e.g. `payment_method` as a parameter.

@mt-clearhaus @soltveit @ad-clearhaus @jsmpereira WDYT?

mt-clearhaus:
> We could require `POST /cards` to have e.g. `payment_method` as a parameter.

I think this makes sense. Maybe have it default to `card`? W.r.t. MPO, we also need to allow the optional `mobilepayonline[phonenumber]` for `POST /cards`, right?

I tend to believe that this calls for having `payment_method` on the authorisation, like I have proposed in https://github.com/clearhaus/almanac/pull/86 ?

ct-clearhaus:
> (...) we also need to allow the optional `mobilepayonline[phonenumber]` for `POST /cards`, right?

True.

I think adding `payment_method` as a separate parameter is not particularly nice. It's lacks "uniformity" IMO. Having `mobilepayonline=true` plus optionally `mobilepayonline[phonenumber]` is perhaps slightly less awful?

> I tend to believe that this calls for having payment_method on the authorisation (...)

Agree.

mt-clearhaus:
> Having `mobilepayonline=true` plus optionally `mobilepayonline[phonenumber]` is perhaps slightly less awful?

Agree.

ct-clearhaus:
After extensive discussions with myself and @jsmpereira (thanks!) ... let me divide.

#### Non-recurring transactions

When Mobile Pay (Online) (MPO) is the source of the payment credentials (being a PAN or a token), then you would always have what it takes to `POST /authorizations?recurring=false`. That is, there's no need to allow `mobilepayonline` when `POST`'ing to `/cards/:id/authorizations?recurring=false`.
The same holds for Apple Pay (AP).

If the PSP `POST /authorizations?recurring=false` using the stored credentials they obtained via MPO or AP, then we cannot know. This can be either an MIT (Merchant Initiated Transaction) or CIT (Cardholder Initiated Transaction). First of all, we cannot decide; we're assuming it's MIT since otherwise it would be possible to use the "full credentials". (As for AP this is an advantage to the PSP/merchant since they're 3-D Secured. As for MPO the PSP/merchant's advantage is slightly less, but when 2FA is required (due to "large enough" amount), then processing as MPO improves the acceptance rate (due to our rules).

#### Recurring transactions

The PSP can either `POST /cards/:id/authorizations?recurring=true` or `POST /authorizations?recurring=true` with the stored credentials they obtained through MPO or AP This is no problem since we track the latest initial recurring and can see if that was created using `mobilepayonline` or `applepay`.

ct-clearhaus:
PAN, account number, card number, ...

What name to use for referring to "either Primary Account Number or a Token (surrogate PAN)"? Somewhere a token is also called a "DPAN"/"Digitised PAN" or a "token PAN". And since PAN can also mean "Payment Account Number" ...
I still dislike `pan` as suggested by this PR, but I have no better suggestion. (Keeping `number` is worse IMO.) So, I think I'll just learn to accept that PAN will mean a "real, non-digitesed/non-tokenised/non-token PAN" or a token. Or, PAN can still mean Primary Account Number, but in `card[pan]` or `mobilepayonline[pan]` you may put a PAN or a token. 🤐

My argumentation:
* Card number: Well, it might be imprinted on no card, not even a virtual card. 👎
* Account number: Nobody would understand what it is without confusion, so we would need everywhere to explain that it's a "PAN or tokenised PAN". 👎
* Number: Yeah, but then "digits" would be equally good. They are numbers. But so are phone numbers. And expiry date. And CSC. And the amount. Only in the context of "card -> number", it makes sense — because it's then named "card number". But ... see above. 👎
* Both Visa's DMSA Tech. Spec and Mastercard's CIS and Auth Manual still names the field PAN.

@mt-clearhaus @jsmpereira WDYT? Suggestions?

mt-clearhaus:
I completely agree with your arguments, especially the bullet about Visa and MasterCard still using the field name PAN.

jsmpereira:
All else being equal one could argue that if the available "literature" (Visa's DMSA Tech. Spec and Mastercard's CIS and Auth Manual, and potentially others) uses `PAN`, that could be a "good enough" reason (because consistency?) to conform with `PAN`. Maybe? 😔

source/index.md
```diff
@@ -342,14 +342,11 @@ them to provide card information for subsequent payments.
...
-tokens. Card tokens are created implicitly when the first [authorization is
-created](#reserve-money) no matter the payment method used.
-
-Actual recurring transactions must be made by
-[creating an authorization](#authorizations).
-
+Many PSPs has a subscription concept for supporting recurring payments.
+The first approved authorization is the initial recurring authorization (also
+known as "first in series"), all later authorizations are called subsequent
+authorizations. Clearhaus support subscriptions in the form of recurring
```
mt-clearhaus: s/support/supports/

source/index.md
```diff
@@ -568,26 +566,76 @@ otherwise, exactly one payment method must be used.
...
-<dl class="dl-horizontal">
-  <dt>applepay[token]</dt>
-  <dd>[:json:] <br /> Full, raw <code>PKPaymentToken</code> object, UTF-8 encoded serialization of a JSON dictionary.</dd>
-  <dt>applepay[symmetric_key]</dt>
-  <dd>[:hex:] <br /> EC symmetric AES key that can decrypt <code>data</code>
-      from the <code>PKPaymentToken</code>; hex formatted as it is a SHA256.</dd>
-</dl>
+Apple Pay requires data of the payment token to be decrypted. For this, the
+merchant's private key is needed to derive a symmetric key which is then used to
+decrypt the PAN, expiry etc.
```
mt-clearhaus: Suggestion: s/to decrypt the PAN, expiry etc/for the decryption/

ct-clearhaus: I like it. What about
```
diff --git a/source/index.md b/source/index.md
index 0ce2c8b..f236698 100644
--- a/source/index.md
+++ b/source/index.md
@@ -566,9 +566,9 @@ method must be used.

 ##### Method: `applepay`

-Apple Pay requires data of the payment token to be decrypted. For this, the
-merchant's private key is needed to derive a symmetric key which is then used to
-decrypt the PAN, expiry etc.
+Apple Pay requires the payment details (PAN, expiry, etc.) of the payment token
+to be decrypted. For this, the merchant's private key is needed to derive a
+symmetric key which is then used for decryption.
```
?

source/index.md
```diff
@@ -568,26 +566,76 @@ otherwise, exactly one payment method must be used.
...
+
+You can either
+
+* send us the private key and the merchant id (either directly or
+  indirectly by sending the merchant's certificate), then we will derive
+  the symmetric key and decrypt the payment details; or
+
+* derive the symmetric key from the payment token's ephemeral public key
+  together with the merchant's private key and merchant id/certificate.
```
mt-clearhaus: Suggestion:

```
You can either send us

* the private key and the merchant ID (either directly in <code>applepay[merchant_id]</code> or indirectly by sending us the merchant's certificate) or
* the symmetric key.

For the former option we derive the symmetric key and decrypt the payment details. For the latter option you need to derive the symmetric key, which is done using the payment token's ephemeral public key, the merchant's private key and the merchant ID.
```

source/index.md
```diff
@@ -568,26 +566,76 @@ otherwise, exactly one payment method must be used.
...
+Your systems must derive the symmetric key. For this, we refer to [our reference
+implementation](https://github.com/clearhaus/pedicel) written in Ruby; see
+[Apple's documentation for the <code>PaymentToken</code>
+object](ApplePay-PaymentToken) for more information.
+
+<dl class="dl-horizontal">
+  <dt>applepay[payment_token]</dt>
+  <dd>[:json:] <br /> Full, raw <code>PKPaymentToken</code> object, UTF-8 encoded serialization of a JSON dictionary.</dd>
+  <dt>applepay[symmetric_key]</dt>
+  <dd>[:hex:] <br /> EC symmetric AES key (unique per transaction) that can
```
mt-clearhaus: s/EC s/S/ ? Since EC in general does not have anything do to with the symmetric en-/decryption, right? The only connection is that the algorithm is AES-256 for `EC_v1` and AES-128 for `RSA_v1`... I guess we will also use `applepay[symmetric_key]` for the latter, if/when we support it?

source/index.md
ct-clearhaus: Will update `s/erchant id/erchant ID/` elsewhere too, and consider writing "Apple Pay merchant ID" where I think it's appropriate (since "merchant ID" is actually both MID in the context of 3DS and also wrt. Apple Pay, but it's two different IDs).

source/index.md
```diff
@@ -568,26 +566,76 @@ otherwise, exactly one payment method must be used.
...
+    decrypt <code>data</code> from the <code>PKPaymentToken</code>; hex
+    formatted as it is a SHA256.</dd>
+</dl>
+
+###### Using the EC private key
+
+When using the EC private key the merchant id is also necessary to decrypt the
```
mt-clearhaus: Suggestions:

s/key the/key, the/
s/merchant id/merchant ID/

source/index.md
ct-clearhaus: C&P error from when it was `applepay[private_key]`.

ct-clearhaus: In fact, by using `applepay[symmetric_key]` we do support `RSA_v1`. We do however only support `EC_v1` when the PSP prefer to send the private key.

source/index.md
```diff
@@ -568,26 +566,76 @@ otherwise, exactly one payment method must be used.
...
+payment token data. This can be extracted from the merchant's certificate.
+
+<dl class="dl-horizontal">
+  <dt>applepay[payment_token]</dt>
+  <dd>[:json:] <br /> Full, raw <code>PKPaymentToken</code> object, UTF-8 encoded serialization of a JSON dictionary.</dd>
+  <dt>applepay[private_key]</dt>
+  <dd>[:PEM:] <br />
+    PEM formatted merchant EC private key. <br />
```
mt-clearhaus: Suggestion: s/PEM formatted/PEM-formatted/ (and elsewhere)

http://www.grammar.cl/english/compound-adjectives.htm 🙂

mt-clearhaus: Likewise for "Hex encoded".

ct-clearhaus: Neither hex nor PEM did indeed format anything 🙂

source/index.md
```diff
@@ -597,11 +645,12 @@ otherwise, exactly one payment method must be used.
...
   <dt>mobilepayonline[expire_year]</dt>
   <dd>[0-9]{4} <br /> Expiry year of card to charge.</dd>
-  <dt>mobilepayonline[phonenumber]</dt>
+  <dt>mobilepayonline[phone_number]</dt>
   <dd>
     [\x20-\x7E]{1,15} <br />
     <i>Optional when transaction is signed and partner is trusted.</i> <br />
-    Phone number from where the PAN originates.
+    Phone number from where the PAN originates. <br />
+    Randomly transactions will be selected and the translation will be verified.
```
mt-clearhaus: Suggestion: Random transactions will be selected for verification.

source/index.md
mt-clearhaus: 👍

source/index.md
ct-clearhaus: I have already pushed this:
> Transactions will be randomly selected and the translation will be verified.

Could I ask you to refresh the page? 🙂

mt-clearhaus: Of couse. I was reviewing a specific commit (https://github.com/clearhaus/gateway-api-docs/pull/53/commits/933f5c3f109160e8ebe5096635ac62f38d813846).

source/index.md
```diff
@@ -602,36 +605,36 @@ object](ApplePay-PaymentToken) for more information.
   <dt>applepay[payment_token]</dt>
   <dd>[:json:] <br /> Full, raw <code>PKPaymentToken</code> object, UTF-8 encoded serialization of a JSON dictionary.</dd>
   <dt>applepay[symmetric_key]</dt>
-  <dd>[:hex:] <br /> EC symmetric AES key (unique per transaction) that can
-    decrypt <code>data</code> from the <code>PKPaymentToken</code>; hex
-    formatted as it is a SHA256.</dd>
+  <dd>[:hex:] <br /> symmetric AES key (unique per transaction) that can
```
mt-clearhaus: s/symmetric/Symmetric/

source/index.md
```diff
@@ -492,22 +520,146 @@ POST https://gateway.clearhaus.com/authorizations
...
+  <b>Notice:</b> An authorization made with <code>applepay</code> cannot be a
+  subsequent recurring authorization.
+</p>
+
+###### Using the symmetric key
+
+Your systems must derive the symmetric key. For this, we refer to [our reference
+implementation](https://github.com/clearhaus/pedicel) written in Ruby; see
+[Apple's documentation for the <code>PaymentToken</code>
+object](ApplePay-PaymentToken) for more information.
```
kse-clearhaus: s/(ApplePay-PaymentToken)/[ApplePay-PaymentToken]/

kse-clearhaus:
Looks good, I do have a few change suggestion for a more cohesive documentation:

- Describe that it is only possible to enable signing of request, when we have exchanged public  keys, which is done by contacting us.
- Ammend every deprecation to either suggest a different method, or to make it clear that the deprecation means loss of functionality.
- I would change the line
  ```
  The account resource holds basic merchant account information. Only HTTP GET is supported for this endpoint.
  ```
  into
  ```
  The account resource contains basic merchant account information. Only HTTP GET is supported for this endpoint.
  ```
  that is, change `holds` to `contains`.

I might have some comments on the `applepay` documentation, but I haven't delved into them, yet.

source/index.md
```diff
@@ -763,10 +931,13 @@ Declined   |  40000 |  General input error
 
 ### Status message
 
-A status message is included in every response when you make a new transaction.
+A status message may be included in the response when you create a new
+transaction.
 The status message can be used for debugging and may include a more specific
 error message.
 
+Status messages should be shown to the merchant.
```
soltveit: I think this highly depends on the message. I can imagine that most merchants don't want to announce e.g. they have hit a sales limit.

ct-clearhaus: The text says "should be shown to the merchant", not "should be shown to the cardholder".

source/index.md
```diff
@@ -372,18 +377,30 @@ Example response (snippet):
...
+   "_embedded": { "card": { "id": "58dabba0-e9ea-4133-8c38-bfa1028c1ed2" } }
 }
 ````
 
-Above can be repeated whenever you need to reserve money and must be followed
-by a capture transaction.
+This should be followed by a capture except when the amount is `0`.
+
+Subsequent authorizations are made similarly, but neither CSC nor PARes (see
+[3-D Secure](#3-d-secure)) would be included. Use the [cards](#cards) endpoint
```
ct-clearhaus: Add deprecation notice for the cards endpoint.

source/index.md
```diff
@@ -372,18 +377,30 @@ Example response (snippet):
...
+or, when approved for CSC-less transactions, the `card` payment method.
+
+When using the `applepay` payment method, the initial recurring transaction is
+made using `applepay[]` whereas subsequent recurring transactions are made using
+either the responded card token or, when approved for CSC-less transactions, the
```
ct-clearhaus: Add deprecation notice for the cards endpoint.

source/index.md
```diff
@@ -462,15 +484,21 @@ Our API offers six different resources:
...
-POST https://gateway.clearhaus.com/cards/:id/authorizations
 POST https://gateway.clearhaus.com/authorizations
+POST https://gateway.clearhaus.com/cards/:id/authorizations # deprecated
 ````
 
+Authorizations can be created using different payment methods:
+`card`, `applepay`, `mobilepayonline`.
+When authorizations are made on a card resource (`/cards/:id/authorizations`,
+deprecated), the payment method is omitted; otherwise, exactly one payment
+method must be used.
```
ct-clearhaus: Turn around. Cc: @ad-clearhaus.

source/index.md
```diff
@@ -462,15 +484,21 @@ Our API offers six different resources:
...
+
 #### Parameters
 
 <dl class="dl-horizontal">
   <dt>amount</dt>
-  <dd>[1-9][0-9]{0,9} <br /> Amount in minor units of given currency (e.g. cents if in Euro).</dd>
+  <dd>[0-9]{1,10} <br /> Amount in minor units of given currency (e.g. cents if in Euro).</dd>
```
ct-clearhaus: `(0|[1-9][0-9]{1,9})`?

source/index.md
```diff
@@ -492,22 +520,146 @@ POST https://gateway.clearhaus.com/authorizations
...
+
+Apple Pay requires the payment details (PAN, expiry, etc.) of the payment token
+to be decrypted. For this, the merchant's private key is needed to derive a
+symmetric key which is then used for decryption.
+
+You can either send us
+
+* the private key and the merchant ID (either directly in
+  `applepay[merchant_id]` or indirectly by sending us the merchant's certificate
+  in `applepay[certificate]`) or
```
ct-clearhaus: Remove this option? And point to `pedicel` and `pedicel-pay`.

source/index.md
```diff
@@ -492,22 +520,146 @@ POST https://gateway.clearhaus.com/authorizations
...
+  <dt>mobilepayonline[pan]</dt>
+  <dd>[0-9]{12,19} <br /> Primary account number of card to charge.</dd>
+  <dt>mobilepayonline[expire_month]</dt>
+  <dd>[0-9]{2} <br /> Expiry month of card to charge.</dd>
+  <dt>mobilepayonline[expire_year]</dt>
+  <dd>[0-9]{4} <br /> Expiry year of card to charge.</dd>
+  <dt>mobilepayonline[phone_number]</dt>
+  <dd>
+    [\x20-\x7E]{1,15} <br />
+    <i>Optional when transaction is signed and partner is trusted.</i> <br />
```
ct-clearhaus: Needs to be signed and trusted anyway because there is no CSC.

source/index.md
```diff
@@ -849,6 +1020,17 @@ https://gateway.clearhaus.com/account
 
 ## Changes
 
+### Add payment methods
+
+Payment methods have been added to the authorization resource. Also, a few
```
ct-clearhaus: `s/authorization resource/authorizations resource/`

source/index.md
```diff
@@ -492,22 +520,146 @@ POST https://gateway.clearhaus.com/authorizations
...
+  <!-- deprecated -->
+  <dt><strike>threed_secure[pares]</strike></dt>
+  <dd>
+    Deprecated! Please use <code>card[pares]</code>. <br />
+    [:base64:] <br />
+    See more information at <a target="_blank" href="http://docs.3dsecure.io">3Dsecure.io</a>.
+  </dd>
+  <dt><strike>card[number]</strike></dt>
+  <dd>
+    Deprecated! Please use <code>card[pan]</code>. <br />
```
ct-clearhaus: `card[number]` should not be deprecated.

source/index.md
```diff
@@ -652,19 +805,30 @@ POST https://gateway.clearhaus.com/cards
...
-  <dd>[0-9]{3} <br /> Card Security Code.</dd>
+  <dd>
+    [0-9]{3} <br />
+    <i>Optional when transaction is signed and partner is trusted.</i> <br />
+    Card Security Code.
+  </dd>
+  <!-- deprecated -->
+  <dt><strike>card[number]</strike></dt>
+  <dd>
+    Deprecated! Please use <code>card[pan]</code>. <br />
```
ct-clearhaus: `card[number]` should not be deprecated.

source/index.md
ct-clearhaus: @ad-clearhaus, will you commit and push your though? 🙂

source/index.md
ct-clearhaus: Don't bother.

✅

source/index.md
ct-clearhaus: ✅

source/index.md
```diff
@@ -462,15 +483,21 @@ Our API offers six different resources:
...
 
 ````
-POST https://gateway.clearhaus.com/cards/:id/authorizations
 POST https://gateway.clearhaus.com/authorizations
+POST https://gateway.clearhaus.com/cards/:id/authorizations # deprecated
 ````
 
+Authorizations can be created using different payment methods:
+`card`, `applepay`, `mobilepayonline`.
+Exactly one payment method must be used, unless the authorization is made on
```
ad-clearhaus: Changed sentence order. Common case first, special case last.

kse-clearhaus:
The length on account `descriptor` should match the `text_on_statement` length.

https://github.com/clearhaus/gateway-api-docs/blame/d6db75e00972ff79b4f89ffc0e97e2bb0583d695/source/index.md#L829-L834

ct-clearhaus:
> The length on account `descriptor` should match the `text_on_statement` length, I assume.

I did think about this divergence, and it's on purpose. It will be a valid configuration of a merchant to have the default descriptor be `x` or even the empty string; the consequence is that the merchant is required to send a `text_on_statement` on the transactions for it to not be rejected due to the ToS requirements ("all same character").

source/index.md
```diff
@@ -462,15 +483,21 @@ Our API offers six different resources:
...
+a card resource(`/cards/:id/authorizations`, deprecated), in which case the
```
mt-clearhaus: Missing space between `resource` and `(`

source/index.md
```diff
@@ -340,27 +342,30 @@ them to provide card information for subsequent payments.
...
+The first approved authorization is the initial recurring authorization (also
+known as "first in series"), all later authorizations are called subsequent
+authorizations. Clearhaus supports subscriptions in the form of recurring
+transactions.
 
 ### Repeatedly reserve money
 
-A recurring payment is made by making an authorization based on a card
-resource and setting `recurring` parameter to `true`:
+A recurring payment is made by making an authorization and setting `recurring`
```
mt-clearhaus: Perhaps we confuse the reader with too many terms for the same thing? In line 349 we have "recurring transactions"; here we have "recurring payment".

mt-clearhaus: @kse-clearhaus  WDYT about `s/payment/transaction/`?

source/index.md
```diff
@@ -479,35 +506,119 @@ POST https://gateway.clearhaus.com/authorizations
...
 
 <p class="alert alert-info">
-<b>Notice:</b> The card dictionary is not required when making an authorization
-based on a card resource. However, <code>card[csc]</code> should be populated
-when CSC is available.
+  <b>Notice:</b> Signing is required to use the <code>applepay</code> payment
+  method.
+  <br />
+  <b>Notice:</b> An authorization made with <code>applepay</code> is
+  strongly authenticated.
```
mt-clearhaus: Perhaps we should explain "strongly authenticated"? This is the first occurrence of the term.

mt-clearhaus: We could add e.g. "(SCA in PSD2)"...

@kse-clearhaus WDYT?

source/index.md
```diff
@@ -849,6 +975,17 @@ https://gateway.clearhaus.com/account
...
+### Add payment methods
+
+Payment methods have been added to the authorizations resource. Also, a few
+parameters and an endpoint have been deprecated. Refer to [the documentation
+source code changes](https://github.com/clearhaus/gateway-api-docs/pull/53/files)
+for the exact documentation change.
+
+Please notice that there is no major version number change, so we stay backwards
+compatible until the deprecations take effect. Expect the deprecations to happen
+within a month or so; it will be announced separately.
```
mt-clearhaus: Perhaps `s/a month/two months/` since we aim for mid-October?

@kse-clearhaus WDYT?

kse-clearhaus: :+1:

source/index.md
```diff
@@ -849,6 +975,17 @@ https://gateway.clearhaus.com/account
 
 ## Changes
 
+### Add payment methods
```
kse-clearhaus: We should add date of deployment here @mt-clearhaus

mt-clearhaus: :+1:

kse-clearhaus:
@mt-clearhaus I agree :+1:

mt-clearhaus:
Deployed on http://docs.gateway.clearhaus.com.

PR #64, opened 2018-08-17
Credits with card details

ct-clearhaus:


source/index.md
```diff
@@ -713,13 +713,53 @@ No parameters are needed to make a new void transaction.
...
 #### Parameters
 
 ````
-POST https://gateway.clearhaus.com/cards/:id/credits
+POST https://gateway.clearhaus.com/credits
+````
+
+<dl class="dl-horizontal">
+  <dt>amount</dt>
+  <dd>[1-9][0-9]{0,9} <br />
```
mt-clearhaus: We use `[0-9]{1,10}` for amount for `/authorizations` now :slightly_smiling_face:

mt-clearhaus: Of course, we will have to adapt https://github.com/clearhaus/klovedal/blob/5f097d212b9f5d87724ab5ceb0c67f2b13e8f7b4/lib/klovedal/request/credit.rb#L8

source/index.md
```diff
@@ -713,13 +713,53 @@ No parameters are needed to make a new void transaction.
...
+    A reference to an external object, such as an order number.
+  </dd>
+  <dt>card[pan]</dt>
+  <dd>[0-9]{12,19} <br /> Primary account number of card to charge.</dd>
+  <dt>card[expire_month]</dt>
+  <dd>[0-9]{2} <br /> Expiry month of card to charge.</dd>
+  <dt>card[expire_year]</dt>
+  <dd>20[0-9]{2} <br /> Expiry year of card to charge.</dd>
+  <dt>card[csc]</dt>
+  <dd>[0-9]{3} <br /> <i>Optional.</i> <br /> Card Security Code.</dd>
```
mt-clearhaus: We need to decide if we want to offer `card[csc]` or not.

source/index.md
ct-clearhaus: "0-credits" should not be allowed; 0-auths are allowed.

source/index.md
ct-clearhaus: Indeed. We've allowed it so far; I guess we can check if we ever processed credits with CSC?

source/index.md
mt-clearhaus: Do'h. Good point.

source/index.md
mt-clearhaus: We have not processed credits with CSC so far.

ct-clearhaus: Let's kill it.

ct-clearhaus: Uhm, or, wait. We should perhaps consider how many `POST /cards?card[csc]=...` was done "just before" the accompanying `POST /cards/:id/credits` ...

ct-clearhaus: Let's keep it for now and remove later if never used.

ct-clearhaus:
@mt-clearhaus Should we include another entry in the "Changes" section?

source/index.md
```diff
@@ -1015,6 +1015,12 @@ https://gateway.clearhaus.com/account
 
 ## Changes
 
+### Add credits resource
+
+Starting 2018-xx-xxTxx:xx:xx+00:00 the `/credits` resource has been added,
+enabling creation of credits without card tokenization. This is an essential
+addition, as card tokenization is deprecated.
```
ct-clearhaus: Uhm, I think the more important note is that `/cards/:id/credits` will also be deprecated — and what the deadline for this functionality is.

source/index.md
```diff
@@ -975,6 +1015,13 @@ https://gateway.clearhaus.com/account
 
 ## Changes
 
+### Add credits resource
+
+Starting 2018-xx-xxTxx:xx:xx+00:00 the `/credits` resource has been added,
```
ct-clearhaus: IMO: `s/Starting 2018-xx-xxTxx:xx:xx+00:00 the/The/` and then we just don't deploy this PR before the implementation is in production.

We shouldn't forget also to communicate this explicitly to partners (that do credits).

mt-clearhaus: We have starting dates for a few other change announcements, so I will keep it. I acknowledge that a date is partly deducible from this PR (which is public), however, I think a changelog needs versions, dates or both in order to be usable on its own.

mt-clearhaus:
Deployed to http://docs.gateway.clearhaus.com :heavy_check_mark:

PR #65, opened 2018-09-03
Credits using slash credits

mt-clearhaus:
@ad-clearhaus pointed out that our payout example still uses `/cards/:id/credits`.


mt-clearhaus:
Deployed to http://docs.gateway.clearhaus.com :heavy_check_mark:

PR #67, opened 2018-11-13
Simplify and exemplify applepay[payment_token]

ct-clearhaus:


absrd:
My point was showing the expected JSON in transit, if one should use that instead urlencoded form data, but that might be a too specific example for your documentation.

source/index.md
```diff
@@ -579,7 +579,10 @@ object][ApplePay-PaymentToken] for more information.
 
 <dl class="dl-horizontal">
   <dt>applepay[payment_token]</dt>
-  <dd>[:json:] <br /> Full, raw <code>PKPaymentToken</code> object, UTF-8 encoded serialization of a JSON dictionary.</dd>
+  <dd>[:json:] <br />
+    Full <code>PKPaymentToken</code> serialized as JSON, supplied as a string.
+    Example: `{"paymentData":{"version":"EC_v1", "data":"<Base64>",...},"paymentMethod":{...},...}`
```
mt-clearhaus: Remove the space between `"EC_v1",` and `"data"`.
Perhaps add a line break between `string.` and `Example:`.

mt-clearhaus: i.e. `<br />` after `string.` :slightly_smiling_face:

PR #68, opened 2018-12-11
Correct status code as transaction amount

mt-clearhaus:
Actually, rewrite the entire section on test card numbers.


mt-clearhaus:
Deployed.

PR #69, opened 2019-01-09
 Add google pay documentation

kse-clearhaus:
The PR is ready to be merged. It is supported in our Gateway with protocol version `ECv2`.


source/index.md
```diff
@@ -600,9 +602,46 @@ object][ApplePay-PaymentToken] for more information.
...
 
-##### Method: `mobilepayonline`
 
-<dl class="dl-horizontal">
+---
+##### **Method**: `googlepay`
+
+To accept a payment using Google Pay, apart from the complete payment method
+token and merchant ID, the derived shared secret is required. This must be
+derived according to the documentation.
```
mt-clearhaus: I would not mention the token and merchant ID, as these are evident from the list below. Two simpler sentences are e.g. "The shared secret must be derived from the ephemeral public key and your private key. The cryptography involved in Google Pay is described here.". If this suggestion is used, I suggest the text "The shared secret derived from the ephemeral public key and your private key, Base64 encoded." in the list below is changed to "The shared secret. Base64 encoded." to avoid repetition.

source/index.md
```diff
@@ -600,9 +602,46 @@ object][ApplePay-PaymentToken] for more information.
...
+
+The cryptography involved is described [here][GooglePay-PaymentCryptography].
+
+<dl>
+  <dt>googlepay[token]</dt>
+  <dd>[:json:] <br /> Raw payment method token as received in response from Google. UTF-8
+    encoded serialization of a JSON dictionary.</dd>
+  <dt>googlepay[shared_secret]</dt>
+  <dd>
```
ct-clearhaus:
```suggestion
  <dd>[:hex:]
```
?

source/index.md
```diff
@@ -600,9 +602,46 @@ object][ApplePay-PaymentToken] for more information.
...
+    The shared secret derived from the ephemeral public key and your private key, Base64 encoded.
```
ct-clearhaus: The description is correct, but just slightly different from the one for `applepay`. I do not have a strong preference as to which to chosen, I just think they should "agree more" as they are very similar under the hood 🙂

source/index.md
```diff
@@ -600,9 +602,46 @@ object][ApplePay-PaymentToken] for more information.
...
+  </dd>
+  <dt>googlepay[merchant_id]</dt>
+  <dd>
+    The ID of the merchant, string matching the format `merchant:<ID>`.
+  </dd>
```
ct-clearhaus: Let's keep the wrapping consistent 🙂

source/index.md
```diff
@@ -600,9 +602,46 @@ object][ApplePay-PaymentToken] for more information.
...
+<dl>
+  <dt>googlepay[token]</dt>
+  <dd>[:json:] <br /> Raw payment method token as received in response from Google. UTF-8
+    encoded serialization of a JSON dictionary.</dd>
+  <dt>googlepay[shared_secret]</dt>
+  <dd>
+    The shared secret derived from the ephemeral public key and your private key, Base64 encoded.
+  </dd>
+  <dt>googlepay[merchant_id]</dt>
+  <dd>
```
ct-clearhaus:
```suggestion
  <dd>`merchant:[:hex:]`
```

But then we should likely have all the regexp-ish format descriptions be `code-formatted` ... dunno how to handle this. Perhaps the textual version below is better, and then just have `<dd>[\x20-\x7E]{10,??}`. WDYT?

Are there any length requirements for the merchant ID btw.?

source/index.md
```diff
@@ -539,9 +539,10 @@ payment method must be omitted.
   </dd>
 </dl>
 
-##### Method: `card`
+---
+##### **Method**: `card`
```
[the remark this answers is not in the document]
kse-clearhaus: I guess this was an attempt at a mini-fix, just to make it more usable.
My guess is I can find some better formatting, if need be.
Getting someone else to do it would be just fine by me :)

source/index.md
```diff
@@ -539,9 +539,10 @@ payment method must be omitted.
...
 
-<dl class="dl-horizontal">
+<dl>
```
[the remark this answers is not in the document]
kse-clearhaus: The width is defined by `bootstrap` and the conversion of the document, we can change it, but it might be hacky.
Besides, we want a limited page width. The problem I think is the formatting of the `<dd>`. The wrapping is done horribly.
Let's just public LaTeX docs! :)

source/index.md
```diff
@@ -600,9 +602,46 @@ object][ApplePay-PaymentToken] for more information.
...
 
-<dl class="dl-horizontal">
+---
+##### **Method**: `googlepay`
+
+To accept a payment using Google Pay, apart from the complete payment method
+token and merchant ID, the derived shared secret is required. This must be
+derived according to the documentation.
+
+The cryptography involved is described [here][GooglePay-PaymentCryptography].
```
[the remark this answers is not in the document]
kse-clearhaus: :+1:
Better.

source/index.md
kse-clearhaus: No, I haven't been able to find any format information for `merchantID`.

source/index.md
kse-clearhaus: Well. They are similar, but the secret passed here is not a  symmetric AES.
Combined with the (Google signed) ephemeral public key it is used to create the symmetric AES key and a MAC key.

I don't know. Thinking of going with something like:
```html5
  <dt>googlepay[shared_secret]</dt>
  <dd>[:base64:]<br />The shared secret derived from the ephemeral public key and
    your private key.</dd>
```

source/index.md
kse-clearhaus: I cannot find any information about the format of merchantID at all, sadly.

Hmm. I don't know. I'll have to try a few things.

kse-clearhaus: The most valid thing would be `merchant:.+`, but it doesn't look that good rendered.

kse-clearhaus: I think I'm renaming it, it's better referred to as `recipient_id`. Which is `merchant:<merchant_id>`.

source/index.md
```diff
@@ -600,9 +602,41 @@ object][ApplePay-PaymentToken] for more information.
...
+
+<dl>
+  <dt>googlepay[token]</dt>
+  <dd>[:json:]<br />Raw payment method token as received in response from Google. UTF-8
+    encoded serialization of a JSON dictionary.</dd>
+  <dt>googlepay[shared_secret]</dt>
+  <dd>[:base64:]<br />The shared secret derived from the ephemeral public key and
+    your private key.</dd>
+  <dt>googlepay[recipient_id]</dt>
+  <dd>merchant:[:merchant_id:]<br />The ID of the merchant preprended with "merchant:".</dd>
```
ct-clearhaus: Oh, wait. Why don't *we* just append that `"merchant:"` string and require the "actual" merchant ID to be given in the API?

source/index.md
```diff
@@ -600,9 +602,41 @@ object][ApplePay-PaymentToken] for more information.
...
+<p class="alert alert-info">
+  <b>Notice:</b> Signing is required to use the <code>googlepay</code> payment
+  method.
+  <br />
+  <b>Notice:</b> An authorization made with <code>googlepay</code> may be fully
+  3-D Secured, 3-D Secure attempted, or with no 3-D Secure; this is indicated by
+  the <code>eciIndicator</code> of the <code>paymentMethodDetails</code>.
+  <br />
+  <b>Notice:</b> An authorization made with <code>googlepay</code> cannot be a
+  subsequent recurring authorization.
```
ct-clearhaus: Perhaps add:

> If marked as `recurring`, the authorization will be an initial recurring authorization.

It would be appropriate to notice that it'd be a "CIT" (Cardholder Initiated Transaction), but I do not thing it's appropriate to add in this PR. It should be done throughout the documentation (thus, in a separate PR) 🙂

source/index.md
ct-clearhaus: Ahh, yes, right. I like your thinking! 👍

source/index.md
ct-clearhaus: I'm somewhat a fan of your renaming suggestion — we already have quite some "things" named "merchant id" 🙁

source/index.md
kse-clearhaus: Yeah. You're absolutely correct. I can't believe I didn't consider this.. :keyboard: :timer_clock: 
Working on it now.

kse-clearhaus: :heavy_check_mark:

ct-clearhaus:
@kse-clearhaus, I took the freedom to push directly; the " Is Google Pay SCA? No, not without 3DS." box was ticked, but I couldn't find the text mentioning it.

source/index.md
```diff
@@ -600,9 +602,50 @@ object][ApplePay-PaymentToken] for more information.
...
+To accept a payment using Google Pay, apart from the complete payment method
+token and merchant ID, the derived shared secret is required.
+Please refer to the [official documentation][GooglePay-PaymentCryptography].
+Supported protocol versions are `ECv1` and `ECv2`.
+
+<dl>
+  <dt>googlepay[token]</dt>
+  <dd>[:json:]<br />Raw payment method token as received in response from Google. UTF-8
+    encoded serialization of a JSON dictionary.</dd>
+  <dt>googlepay[shared_secret]</dt>
```
tjconcept: Minor detail, but Google is calling this a "shared key".

ct-clearhaus: @kse-clearhaus shouldn't we just update the API (and the implementation) accordingly? Or was there a reason for the current naming?

ct-clearhaus:
Public notice: Perhaps we'll end up throwing in 3DSv2 stuff for Google Pay (and thus updating this PR) before merging. Still undecided.

Notice to @clearhaus/pci: I guess we should keep this PR aligned with the Beta version available in `gateway.(test.)clearhaus.com`; that is, changes to this PR should probably be done as PRs onto this rather than "direct" pushes. Then, when we're out of Beta, this PR will be merged.

source/index.md
```diff
@@ -600,9 +602,50 @@ object][ApplePay-PaymentToken] for more information.
...
+  <dd>[:base64:]<br />The shared secret derived from the ephemeral public key and
+    your private key.</dd>
+  <dt>googlepay[merchant_id]</dt>
+  <dd>[\x21-\x7E]+
+    <i><a target="_blank" href="http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">ASCII printable characters exluding space</a></i>
```
ct-clearhaus:
```suggestion
    <i><a target="_blank" href="http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">ASCII printable characters excluding space</a></i>
```

source/index.md
```diff
@@ -600,9 +602,50 @@ object][ApplePay-PaymentToken] for more information.
...
+  <br />
+  <b>Notice:</b> An authorization made with <code>googlepay</code> may be fully
+  3-D Secured, 3-D Secure attempted, or with no 3-D Secure; this is indicated by
+  the <code>eciIndicator</code> of the <code>paymentMethodDetails</code>.
+  <br />
+  <b>Notice:</b> An authorization made with <code>googlepay</code> is
+  strongly authenticated (SCA in PSD2) if fully 3-D Secure.
+  <br />
+  <b>Notice:</b> An authorization made with <code>googlepay</code> cannot be a
+  subsequent recurring authorization.
```
ct-clearhaus:
```suggestion
  subsequent recurring authorization. If marked as <code>recurring</code>, the
  authorization will be an initial recurring authorization.
```

source/index.md
ct-clearhaus:
```suggestion
  <dt>googlepay[shared_key]</dt>
```

ct-clearhaus (dismissed):

source/index.md
ct-clearhaus: Before committing the suggestion, please notice https://github.com/clearhaus/gateway-api-docs/pull/69#issuecomment-578094920 🙂

kse-clearhaus: Yeah, they likely changed it from version 1 to 2. I remember stressing about the names.

Should we change it? Hmm.

ct-clearhaus: I think we should basically just use whatever Google names it to require as little translation as possible. It's really to cheap to change it now, so I'd say that we should just change it right away; I'll commit it in this PR, and we can simply adjust this PR back if necessary. Thereby, I'll resolve this conversation.

source/index.md
kse-clearhaus: This has been fixed by mentioning it earlier.

source/index.html.md
```diff
@@ -387,10 +387,10 @@ included. For 3-D Secure version 1 (see [3-D Secure](#3-d-secure)), PARes would
 not be included. 3-D Secure version 2 has some support for subsequent
 authorizations.
 
-An initial recurring authorization can also be made using the `applepay` and
-`mobilepayonline` payment methods; subsequent recurring payments, however, must
-be made using the `card` payment method using the card details of the initial
-recurring authorization.
+An initial recurring authorization can also be made using the `applepay`,
+`googlepay` and `mobilepayonline` payment methods; subsequent recurring
```
ct-clearhaus, line 391:
```suggestion
`googlepay` or `mobilepayonline` payment methods; subsequent recurring
```

source/index.html.md
```diff
@@ -528,7 +528,7 @@ POST https://gateway.clearhaus.com/authorizations
 ````
 
 Authorizations can be created using different payment methods:
-`card`, `applepay`, `mobilepayonline`.
+`card`, `applepay`, `googlepay`, `mobilepayonline`.
```
ct-clearhaus, line 531: The ordering here aligns with the ordering below, but any reason for squeezing `googlepay` in (signaling some kind of ordering I don't understand) rather than adding it to the end?

source/index.html.md
```diff
@@ -679,6 +678,57 @@ object][ApplePay-PaymentToken] for more information.
...
+  </dd>
+  <dt>googlepay[shared_key]
+    <span class="type">[:base64:]</span>
+  </dt>
+  <dd>
+    The shared secret derived from the ephemeral public key and your private
+    key.
+  </dd>
+  <dt>googlepay[recipient_id]
+    <span class="type">[\x20-\x7E]+
```
ct-clearhaus, line 705: Since space is not allowed:
```suggestion
    <span class="type">[\x21-\x7E]+
```

But also, isn't there an upper limit on the length?

I went through the documentation I have at hand and couldn't find any.

source/index.html.md
```diff
@@ -679,6 +678,57 @@ object][ApplePay-PaymentToken] for more information.
...
+      <a target="_blank" href="http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">ASCII printable characters excluding space</a>
```
ct-clearhaus, line 706:
```suggestion
      <a target="_blank" href="http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">ASCII printable characters</a> excluding space
```

🤷‍♂️

source/index.html.md
```diff
@@ -679,6 +678,57 @@ object][ApplePay-PaymentToken] for more information.
...
+  </dd>
+</dl>
+
+<p class="alert alert-info">
+  <b>Notice:</b> Signing is required to use the <code>googlepay</code> payment
+  method.
+  <br />
+  <b>Notice:</b> An authorization made with <code>googlepay</code> may be fully
+  3-D Secured, 3-D Secure attempted, or with no 3-D Secure; this is indicated by
+  the <code>eciIndicator</code> of the <code>paymentMethodDetails</code>.
```
ct-clearhaus, line 720: According to https://developers.google.com/pay/api/web/guides/resources/payment-data-cryptography#cryptogram_3ds, `eciIndicator` (RAS Syndrome) is only present for Visa.

source/index.html.md
```diff
@@ -679,6 +678,57 @@ object][ApplePay-PaymentToken] for more information.
...
+  <br />
+  <b>Notice:</b> An authorization made with <code>googlepay</code> is
+  strongly authenticated (SCA in PSD2) if fully 3-D Secure.
```
ct-clearhaus, line 723: I think we need to relax this statement. As a random example, Mastercard LI `kA` has status `Y` and is not SCA.

source/index.html.md
kse-clearhaus: Just phonetically I liked the two `*pay` words to follow each other :)

source/index.html.md
kse-clearhaus: I think you're better equipped to answer that than me :)

source/index.html.md
kse-clearhaus: I'm actually not sure that that is true, _but_, it may be true.

source/index.html.md
```diff
@@ -787,6 +787,59 @@ object][ApplePay-PaymentToken] for more information.
...
+
+<p class="alert alert-info">
+  <b>Notice:</b> Signing is required to use the <code>googlepay</code> payment
+  method.
+  <br />
+  <b>Notice:</b> An authorization made with <code>googlepay</code> is strongly
+  authenticated (SCA in PSD2) if <code>authMethod</code> is
+  <code>CRYPTOGRAM_3DS</code> and the Google Pay guidelines for
+  <i><a href="https://developers.google.com/pay/api/web/guides/resources/sca">sca</a></i>
+  have been followed.
```
ct-clearhaus, line 831: Just a slight "link improvement":
```suggestion
  <b>Notice:</b> An authorization made with <code>googlepay</code> is strongly
  authenticated (SCA in PSD2) if <code>authMethod</code> is
  <code>CRYPTOGRAM_3DS</code> and
  <a href="https://developers.google.com/pay/api/web/guides/resources/sca">the
  Google Pay guidelines for SCA</a> have been followed.
```

What's the reason for the italic?

source/index.html.md
```diff
@@ -787,6 +787,59 @@ object][ApplePay-PaymentToken] for more information.
...
+
+  If <code>authMethod</code> is <code>PAN_ONLY</code>, a 3-D Secure flow is
+  required for SCA.
```
ct-clearhaus, line 834:
```suggestion
  If <code>authMethod</code> is <code>PAN_ONLY</code>, a 3-D Secure flow is
  necessary to obtain strong authentication.
```

Dunno which version is easier to understand correctly 🤷‍♂️

source/index.html.md
```diff
@@ -787,6 +787,59 @@ object][ApplePay-PaymentToken] for more information.
...
+  <br />
+  <b>Notice:</b> An authorization made with <code>googlepay</code> cannot be a
+  subsequent recurring authorization.
+  <br />
+  <b>Notice:</b> The merchant ID for the <code>googlepay</code> test
+  environment is <code>12345678901234567890</code>.
```
ct-clearhaus, line 840:
```suggestion
  <b>Notice:</b> The <code>recipient_id</code> for the <code>googlepay</code>
  test environment is <code>merchant:12345678901234567890</code>.
```

This is correct, right? If so, I think the added detail could perhaps be helpful 🙂

ct-clearhaus (approved):
Only minor suggestions. Looks good 👍

source/index.html.md
kse-clearhaus: The link doesn't look different from ordinary text, so I wanted to highlight it.

It's probably better to make it explicit with a link that looks like `[link]`. I'll do that.

source/index.html.md
kse-clearhaus: I'd prefer to use "SCA", as it's probably easily recognizable as SCA related (and briefer).
But I have no strong opinion. You're welcome to merge if you prefer your edition.

source/index.html.md
kse-clearhaus: Ah, good catch :+1: A remnant of an old version.

PR #71, opened 2019-03-27
Add stop request status codes

ad-clearhaus:


source/index.md
```diff
@@ -926,6 +926,9 @@ Declined   |  40000 |  General input error
            |  40413 |  Insufficient funds
            |  40414 |  Suspected fraud
            |  40415 |  Amount limit exceeded
+           |  40420 |  R0 Stop request. Stop this subscription on this merchant
```
soltveit: Will it confuse if we start referring to scheme response codes (i.e. `R0-R2`)? Do we to mention e.g. `R0`?

mt-clearhaus: Looking at the existing entries in this table, I think we want the text in the Meaning column to match what the gateway returns, i.e. what we have in `.../transaction/status.rb` in the database layer. The Rx response codes are not mentioned in the messages returned by the gateway—at least in the current version of the database layer pull request.

ad-clearhaus: Direct scheme response codes has been removed. R0, R1, R3 are Visa specific, so this solution would leave Mastercard unresolved. Instead are more general concept of **stop** should be adapted.

source/index.html.md
```diff
@@ -963,6 +964,23 @@ Examples:
...
+On any recurring transaction, issuers can respond with <i>stop recurring</i>,
+withdrawing the consent from the cardholder for a merchant to initiate
+transactions on their behalf.
+
+A <i>stop recurring</i> status is both a decline of the current authorization,
+but also a notice to halt all future recurring transactions on this card from
+this merchant.
+
+If <i>stop recurring</i> responses are not adhered to and recurring
+transactions on this card continues, a fee pr. transaction is incurred by
```
kse-clearhaus: Should we mention the size of the fee?

source/index.html.md
```diff
@@ -963,6 +964,23 @@ Examples:
...
+the merchant from the card scheme.
+
+Merchants should contact the cardholder to resolve this issue. Only after the
+withdrawal of consent has been lifted, can recurring transactions be resumed.
```
kse-clearhaus: I would imagine that the cardholder would have to consent a second time, i.e. create a new first recurring with CSC, 3d-secure and so on?

source/index.html.md
ad-clearhaus: No. This might change. And very fast gets very complex when you factor in the different scenarios where a fee can be incurred. The fees for Mastercard and Visa are also not the same.

source/index.html.md
ad-clearhaus: The stop services does not distinguish between different subscriptions.
Getting the customer to agree to a new subscription won't help. The stop-payment blockade will have to be lifted before recurring transaction can be resumed. To quote Visa:

>  **Clearhaus**: What best practices can be pass on to our merchant regarding trying to adhere
to R1 stop requests? On a stop request should merchants freeze all payments,
and contact the cardholder to reaffirm the ones that should continue?
> **Visa**: Yes, exactly!

After the merchant contact the customer/issuer to get the issue resolved, the payments can be resumed. But in reality, this might entail getting the customer to sign-up for at new subscription.

source/index.html.md
```diff
@@ -966,21 +966,20 @@ Examples:
 
 ### Stop recurring status
 
-On any recurring transaction issures have the ability to issue a <i>stop
-recurring</i>, canceling the concent of the card holder, towards a merchant, to
-initiate transactions on their behalf.
+On any recurring transaction, issuers can respond with <i>stop recurring</i>,
```
ad-clearhaus: Incorrect. Stop is something scheme does, not the issuer. That being said, I acknowledge my very long and complicated sentence.

source/index.html.md
tjconcept: Reading this section I was even thinking: why get into scheme politics at all? Shouldn't this documentation stick to protocol and leave the rest to contracts?

ad-clearhaus: Fair point... Thinking.

ad-clearhaus:
Moved "Stop recurring status" section to https://gitlab.com/clearhaus/partner/gateway-api-guideline/

source/index.html.md
```diff
@@ -964,23 +964,6 @@ Examples:
...
-but also a notice to halt all future recurring transactions on this card from
-this merchant.
-
-If <i>stop recurring</i> responses are not adhered to and recurring
-transactions on this card continues, a fee pr. transaction is incurred by
-the merchant from the card scheme.
-
-Merchants should contact the cardholder to resolve this issue. Only after the
-withdrawal of consent has been lifted, can recurring transactions be resumed.
-
```
tjconcept: Couldn't *some* of this be considered protocol and spared? I'm specifically thinking about the two first paragraphs detailing the intent. In other words: true and accurate for all schemes/issuers regardless of contracts.

ad-clearhaus: Only including the first 2 paragraph leaves the obvious question.

*What can I do to "unban" my merchant?*

The last paragraph could be included to calm the nerves a bit

>On any recurring transaction, issuers can respond with stop recurring, withdrawing the consent from the cardholder for a merchant to initiate transactions on their behalf.
>
>A stop recurring status is both a decline of the current authorization, but also a notice to halt all future recurring transactions on this card from this merchant.
>
>Merchants should contact the cardholder to resolve this issue. Only after the withdrawal of consent has been lifted, can recurring transactions be resumed.

tjconcept: You're right.

I can't get the wording right, but I'd rather go with something along the lines of: this is a permanent condition of the card which requires an action from the issuer or cardholder to change.

The original one is probably OK as well.

ad-clearhaus: I have tried to make the section even less opinionated.
But kept emphasis on recurring are disallow until `stop recurring` is lifted.

ad-clearhaus:
The usecase for *stop-request* is recurring transactions, but we do see cases of non-recurring transactions getting stopped by *stop-request* https://github.com/clearhaus/issues-pci/issues/969. Wording has been changed to reflect this.

ad-clearhaus:
After at clear answer from Visa in https://github.com/clearhaus/issues-pci/issues/969 and internal discussion. Wording has changed yet again.

ct-clearhaus:
I will be surprised if we do not (eventually) receive `R[0-3]` for auths that are *not* CoF. In addition, I would like to stay pretty close to what schemes names the things. Thus, I like the thing to be called "something including the word `stop`".

ad-clearhaus:
> I will be surprised if we do not (eventually) receive R[0-3] for auths that are not CoF

We thought so as well, which spawned https://github.com/clearhaus/issues-pci/issues/969
There has been a lot of back and forth on this front. Visa is very inconsistent in its documentations in regards to what a stop-request can stop. But according to Visa, only _recurring_, _instalment_ and _other CoF_ can be blocked. This has been shorted to CoF in our context. AFAICT CoF is the smallest set containing all subsets of stoppable authorizations.

> I like the thing to be called "something including the word stop".

@soltveit  brought up a valid point IRL. The keyword **stop** is a very commandeering and general word and may lead some readers to overreact to this status. E.g. stop to much.
Going with **stop recurring** is technically incorrect, although not for 99%.
Instead we fell back to Visa own meaning  

Screenshot from 2019-05-13 10-40-12

~p. 25, visa-europe-payment-stop-service-acquirer-user-guide.pdf

Furthermore. It also seems to me that Visa changed their minds half way through deploying **stop request**, and are putting less emphasis on the stop wording.
In VAU it is called **stop request**. The service is called Visa European Payment Stop Service. But Visa BASE I, Visa BASE II, Omnipay specification, and Mastercard specification doesn't refer to R1, R3 as stop request.
Mastercard calls it Recurring Payment Cancellation Service btw.

ad-clearhaus:
In the U.S. the service is called Preauthorized Payment Cancellation Service (PPCS). Again, no emphasis on stop.

The hole Visa European Payment Stop Service was original Visas way to comply with a U.K. law, requiring a way for issuers to forcefully stop a subscription. This might later have been polished into the PPCS service and the terms have had some more time to mature.

ct-clearhaus:
After @ad-clearhaus' enlightening of me IRL yesterday + some thought: Since the merchant should call the cardholder when receiving the first `R[0-3]`, this response code *is* indeed pretty commandeering — since the merchant should indeed
> stop/halt/cease/... auths until further actions has been taken (for this PAN).

I do not share @soltveit's fear for merchants/partners "stopping too much"; I am confident that merchants will understand the "for this PAN" part even implicitly. I cannot imagine a merchant/partner would e.g. stop all auths just because they receive one specific response code for one card.

ad-clearhaus:
We had two *stop* options on the table:

> stop recurring

This is technically not enough because any MIT could be stopped.

> stop payment

Both Visa and Mastercard leans toward using payment when describing subscriptions+CoF. Visa European **Payment** Stop Service. Master Preauthorized **Payment** Cancellation Service. But using *stop payment* felt to general.

soltveit:
IIRC from the canteen meeting @ad mentioned that `R0` indicates that the merchant can try again "soon", so commanding them to stop seems wrong ("stop" vs. "try later"). But perhaps these are very rare?

My thinking was also more that `Stop ...` is not telling **why** the transaction got declined. I know many status code are not like that, but we know these general status codes is generating support cases (e.g. `delicned_by_issuer...`). `Card expired` is much better.

Regarding commanding in general:
1) you risk annoying good customers which is the majority
2) you don't stop the bad customers any way sine they will ignore it
3) it make us look like most other acquirers

Merchants probably don't get offended by a status code, but one never know in times like these  😄

ad-clearhaus:
`R0` are rare. Visa European Payment Stop Service(VEPSS) is unable to use `R0`. The ones we see must be from Preauthorized Payment Cancellation Service (PPCS), the US only service.

ad-clearhaus:
I like your point about avoiding commandeering language if possible. I also think that `consent withdrawn` is a more explanatory message.

The best status message is the one with the shortest mental distance from our wording, to the different payment cancellations services of scheme. The less a merchant/PSP has to think about the message to deduce that this is related to VEPSS, PPCS or RPCS the better.

ct-clearhaus:
Sorry in advance. We'll never end up deciding anything if we continue like this 😛

We want the response code to indicate *why*. But there can be multiple reasons. This is likely why we cannot really find something much better than `consent_withdrawn` — but on the other hand, that also doesn't really tell why ... I mean ...
* Clearhaus: "Declined"
* Merchant: "Oh, why?"
* Clearhaus: "Consent withdrawn"
* Merchant "*Why* is *what* consent by *who* withdrawn"?

I do appreciate the thoughts and the intention, but I'm not convinced.

What about:
* `contact_cardholder` — doesn't indicate "why" but at least "what to do". (No, I'm not a fan.)
* Perhaps `contact_cardholder_which_in_turn_will_need_to_contact_the_issuer` would be more correct, but the description following the response code should be enough.
* `isssuer_will_not_allow_the_cardholder_to_buy_your_product_or_service_or_cardholder_does_not_want_your_product_or_service` 😛No, really, how can we communicate this is 1-3 words?
* `unwanted_product_or_service` is perhaps somewhat reasonable, but 4 words and a little too "clashy" I'd say, and also "unwanted by who?"

👀☁️
Stripe: https://stripe.com/docs/declines/codes
`stop_payment_order`
* The card has been declined for an unknown reason.
* The customer should contact their card issuer for more information.

Braintree: https://articles.braintreepayments.com/control-panel/transactions/declines
`2017´
* Cardholder Stopped Billing
* The customer requested a cancellation of a single transaction – reach out to them for more information.
* Hard

`2018`
* Cardholder Stopped All Billing
* The customer requested the cancellation of a recurring transaction or subscription – reach out to them for more information.
* Hard

ad-clearhaus:
Note:
Stripe: https://stripe.com/docs/declines/codes

|   |                   |  |  |
|---|-------------------|---------------------------------------------------|-------------------------------------------------------------------|
|R0 |stop_payment_order	|The card has been declined for an unknown reason.	|The customer should contact their card issuer for more information.|
|R1 |revocation_of_authorization	|The card has been declined for an unknown reason.	|The customer should contact their card issuer for more information.|
|R3 |revocation_of_all_authorizations	|The card has been declined for an unknown reason.	|The customer should contact their card issuer for more information.|

As R0 is US only, our focus is R1 and R3.

ct-clearhaus:
`merchant_blocked_by_cardholder` 💣

ad-clearhaus:
My 2 cents on `merchant_blocked_by_cardholder`

 * It is non-commandeering
 * Has a good **why** factor

Taking a :helicopter: perspective.
The better **why** factor will generate fewer support-cases from confused merchants. Very important!
Actively stopping/cancelling "`merchant_blocked_by_cardholder`" subscriptions should either done by PSP's or something a PSP merchant interface should explain to the merchant how to do. Cancelling a subscription will never be done through `dashboard`. 
This implies that PSP's must have read and understood all our documentation on `merchant_blocked_by_cardholder` to code the tools needed to handle it appropriately. Trying to condense all the knowledge a PSP needs to acquire to understand the implications of **stop request** into a 4 letter status message is a bridge to far to begin with.

@soltveit can you get behind `merchant_blocked_by_cardholder`?

tjconcept:
Could one say that "by cardholder" is not correct because their issuer did it (on their behalf?) in an automated fashion? How about `merchant_blocked_for_cardholder`? I don't know my opinion, just tossing in thoughts.

soltveit:
Agree with @ad-clearhaus's arguments. `merchant_blocked_by_cardholder` is the best suggestion so far.

I also liked `Cardholder (all) stopped billing` (from Braintree) but `merchant_blocked_by_cardholder` better tells what the cardholder actually did (I think). However, I'm struggling a little bit with understanding the difference between `R1` and `R3`. `merchant_blocked_by_cardholder` seems a bit definitive but is that the case with `R1`?

ad-clearhaus:
**tjconcept**
> How about merchant_blocked_for_cardholder?

I prefer `_by_`. It emphasises merchant-cardholder discord, which is easiest solved by merchant-cardholder interaction. I would also hate if a merchant+issuer went over my head and either blocked or lifted a blockade on my behalf. My issuer blockading all gambling merchants on my behalf (due to gambling addiction) seems like a fringe case - Issuers already have tools to solve these issues.


**soltveit**
> However, I'm struggling a little bit with understanding the difference between R1 and R3.

 * **R1**: Block all merchant-initiated-transactions (MIT) from a specific merchant
 * **R3**: Block all MIT from all merchant

Visa gives each individual merchant 3 tries to get it right. So we can condense **R3** and **R1** into one, because our merchants will always have had 3 tries to get it right nomatter if it was a **R1** or **R3**.
To understand why **R3** exists, think about it from the issuers perspective. "How do we block all MIT for our  cardholder without knowing all possible MID?". **R3** is the solution.

Mastercard does not have a **R3** concept. Instead an issuer must after the fact extract a MID and block future authorizations.

ad-clearhaus:
`merchant_blocked_for_cardholder` won out.
This changes the wording of the supplementary explanation to have emphasis on authorization being blocked (which they are), rather than the softer sounding "consent withdrawn".

Issuers has also been removed from the equation. It is true that **R1**,**R3** will be initialized by the issuer and scheme is the one actually catching and declining. But **R1** **R3** is always made on the request of the cardholder. And getting the block lifted should always be a cardholder-merchant interaction. Merchants phoning up banks to get them to white-list withdraws from their customers seems unrealistic. So I think de-emphasizing issuers is the right call.

We _could_ leave out the 2 first section of the paragraph and only write:

> A _merchant blocked by cardholder_ status code is both a decline of the current authorization, but also a notice to **halt all future credentials-on-file authorizations** on this card from this merchant. This includes, but is not limited to, recurring transactions.
>	Only after the blockade has been lifted, by actions taken by the cardholder, may transactions be resumed.

I prefer it as it is.

ct-clearhaus:
> (...) **halt all future credentials-on-file authorizations** (...)

Permaybehaps `s/credentials-on-file/merchant-initiated/`? If a CIT uses CoF, it should not receive `R[0-3]`, right?

I think it's a good call to shorten the description. Explaining CoF, MIT, CIT etc. is kinda out of scope of the docs — I think it belongs in the guideline.

ad-clearhaus:
I would love to use MIT or a number of other short-hand/concepts to make the paragraph as concise as possible. The main issue is that the API docs has very few concepts in it. So suddenly introducing COT/MIT/CoF is not very friendly and inconsistent.

See https://github.com/clearhaus/issues-pci/issues/969#issuecomment-488572998
Screenshot from 2019-05-29 09-31-00

CoF can actually be blocked. But I am all for ignoring this and just bundle it has merchant-initiated-transactions(MIT).

ad-clearhaus:
:+1: :+1: about shortening. It will end up being a very spartan supplementary explanation, but as long as it is clear, that option in preferable.

source/index.html.md
```diff
@@ -963,6 +964,22 @@ Examples:
 }
 ````
 
+### Merchant blocked by cardholder status
+
+Merchants can use credentials-on-file to initiate transactions without the
+cardholder being present to facilitate payment flow. This includes, but is not
+limited to, recurring transactions.
+
+To ensure that money is not withdrawn without the cardholders consent,
```
mt-clearhaus: cardholder's

mt-clearhaus: Perhaps "without a cardholder's consent, the cardholder can"?

source/index.html.md
```diff
@@ -963,6 +964,22 @@ Examples:
...
+a cardholder can block merchants from using credentials-on-file transactions
+on their card.
```
mt-clearhaus: "the card" or "his/her card", I think.

ad-clearhaus:
I would like a review/approval before merging. :eyes:

source/index.html.md
```diff
@@ -965,6 +966,16 @@ Examples:
 }
 ````
 
+### Merchant blocked by cardholder status
+
+A _merchant blocked by cardholder_ status code is both a decline of the current
+authorization, but also a notice to **halt all future
+merchant-initiated-transactions** on this card from this merchant. This
```
mt-clearhaus: `s/d-t/d t/`

source/index.html.md
```diff
@@ -1069,6 +1080,12 @@ Transactions in CLP or UGX will be declined between 2017-10-12T19:00:00+00:00
 and 2017-10-15T19:00:00+00:00 (both inclusive); before this timespan, the
 exponent is 2; after the timespan, the exponent is 0.
 
+### Add merchant blocked by cardholder status
+
+Starting 2019-04-11 a new status code `40420 - Merchant blocked by cardholder`
+is avaliable. Please be advised that appropriate action must be taken to
```
mt-clearhaus: `s/is available/has been added/` ?

mt-clearhaus:
I have resolved my comments. Improvements can be made in a future PR.

PR #72, opened 2019-04-09
Update to stretch

ad-clearhaus:
Did some upgrading due to Debian Jessie mirror becoming unavailable:
 * debian jessie mirrors in `ruby:2.1` no longer active. Upgrade base OS to stretch
 * `npm` not available in `ruby:2.3` (`debian:stretch`), move to `node` images instead, that has `nodeJS` included
 * Update Gemfile to use `ruby:2.3`
 * Simplify Gemfile by removing specific gem versions
 * Move away from `git` gems
 * Deal with changes in `middleman` structure

I have tested this by building and viewing locally. Heroku not tested. But @lj is working on moving to S3 hosting anyway


Dockerfile
```diff
@@ -1,7 +1,7 @@
-FROM ruby:2.1
+FROM node:stretch
```
mt-clearhaus: I would probably have chosen `debian:stretch` as we use that for almost all other images, but hey :man_shrugging: :slightly_smiling_face:

mt-clearhaus: Let me continue the nagging by saying "do'h" to the branch name :smile: 

```
$ git fetch origin ruby2.1->node.stretch
fatal: Couldn't find remote ref ruby2.1-
```

mt-clearhaus: Fair... Bullet 2... Difficult/annoying/time-consuming to install `node` in `debian:stretch`.

PR #76, opened 2019-07-10
Introduce card[3dsv2]

ct-clearhaus:


source/index.html.md
```diff
@@ -567,6 +582,11 @@ payment method must be omitted.
   <br />
   <b>Notice:</b> An authorization that includes <code>card[pares]</code> and/or
   <code>card[csc]</code> cannot be a subsequent recurring authorization.
+  <br />
+  <b>Notice:</b> An authorization made with <code>card[3dsv2]</code> is not a
+  <code>signed 3dsecure</code> transaction; only transactions made with
```
mt-clearhaus: Suggestion: `s/<code>signed 3dsecure<\/code>/cryptographically signed 3D Secure/`

ct-clearhaus: I've tried to reword it.

Your comment also points to the fact that the description should be "specifying", not point to the error message you would see if a 40200 rule violation happens. This is absolutely true!

PR #77, opened 2019-07-10
Introduce card[rreq]

ct-clearhaus:


ct-clearhaus:
Adding `card[ares]` is a little too annoying, I think, so I think this PR should be moved to document `card[3dsv2][rreq]` and `card[3dsv2][ares]`.

It should be explained when to use an ARes and when to use an RReq.

mt-clearhaus:
The PR title is currently misleading, I think. Perhaps rename to "Introduce card[3dsv2][rreq] and card[3dsv2][ares]"?

PR #78, opened 2019-07-10
Add payment method storedcredential

ct-clearhaus:


ct-clearhaus:
`store`*`d`*`credential` indicates that it is not to be used when *being* stored. Also card schemes call it "credential on file" (CoF). So perhaps a better name would be `credentialonfile`? 🙁

ad-clearhaus:
`credentialonfile >> storedcredentials`.
When possible avoid adding/renaming terminology.

PR #83, opened 2019-08-15
Rework credential on file with cardholder/spitballing

kse-clearhaus:


source/index.html.md
```diff
@@ -355,12 +372,20 @@ ad-hoc.
...
+### Subsequent Credential on file
+
+First in series transactions can also be made using the `applepay` and
+`mobilepayonline` payment methods; however, subsequent merchant-initiated
+transactions must be made using the `card` payment method using the card
+details of the initial recurring authorization.
+
+Subsequent authorizations are made by using
+`credential_on_file[previous]=<ID>`, where `<ID>` is the `id` of the previous
+transaction in the series.
```
ct-clearhaus: I've documented it as the ("full") reference, like `/authorizations/:id`, and I have a reason for that. (Hint: "abstract transaction".) 🙂

ct-clearhaus:
The subsection is in the "Examples" section, so I think we should really try to keep it short and "just an example" rather than a full-ish detailed description. The details should go to the reference section IMO.

Perhaps we should completely avoid mentioning UCOF and CIT-only COF in the Examples section and just have a recurring example?

source/index.html.md
mt-clearhaus: At least I think we need to be a bit more precise than "the previous transaction". Is it the authorization or the capture for the transaction? Also, in my opinion we should mention an age limit, e.g. 13 months.

kse-clearhaus: @ct-clearhaus abstract transaction as in?

I agree that we need to be more precise. But we should also be terse.
I'll see what I can do.

kse-clearhaus: But I guess `previous` depends on how we accept continued transactions in our API.

ct-clearhaus: "Abstract transaction": https://github.com/clearhaus/issues-pci/issues/1071#issuecomment-517264132

PR #92, opened 2020-01-07
Make signing of POST requests mandatory

mt-clearhaus:
Prompted by https://github.com/clearhaus/issues-pci/issues/1321#issuecomment-571521073.

### DIFF
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -80,7 +80,7 @@ Number  Text
 POST requests must be signed.
 
 The signature is an RSA signature of the HTTP body; it is represented in hex.
-The signee must be identified by the signing API-key. Both should be provided in
+The signee must be identified by the signing API key. Both should be provided in
 a `Signature` header together with `RS256-hex`:
 
 ```
@@ -97,7 +97,7 @@ signature, otherwise the transaction will fail.
 The RSA signature is an RSASSA-PKCS1-v1_5 signature of the body. It is
 represented in hex.
 
-If the signing API-key is `4390aec7-f76a-4c2f-8597-c87c2d06cb4f`, the signing
+If the signing API key is `4390aec7-f76a-4c2f-8597-c87c2d06cb4f`, the signing
 private key (in PEM format) is
 
 ```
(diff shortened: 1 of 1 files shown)


kse-clearhaus:
We should maybe update the examples we have, since requests must be signed?

mt-clearhaus:
> We should maybe update the examples we have, since requests must be signed?

Good point! :+1: I'll take a stab later today or tomorrow.

source/index.html.md
```diff
@@ -77,10 +77,10 @@ Number  Text
 
 ## Signing requests
 
-Requests can optionally be signed.
+POST requests must be signed.
```
tjconcept: Shouldn't you also state that other requests *can* be signed? If I didn't knew otherwise, I would have a brief second thought before signing every request.

mt-clearhaus: I think the only non-POST request is `GET /account`. This request has an empty body, so one would always be signing the empty string. The gateway accepts a `Signature` header in the `GET /account` request, however, it is ignored.

Am I missing something?

mt-clearhaus:
@clearhaus/pci Note that `POST /authorizations/:id:/captures` requests often have an empty body. I guess we can decide to a) accept this, b) exempt requests with empty bodies from the mandate, or c) make the signing include the resource path (`/authorizations/:id:/captures`). WDYT?

source/index.html.md
tjconcept: I think that's good to know. I simply didn't special-case neither POST nor empty bodies (I remember asking back then to be sure), so I'm signing regardless, I think that is the nicest as a consumer.

ct-clearhaus:
> WDYT?

Should add some nonce (and we should then check for uniqueness to avoid replay).

source/index.html.md
mt-clearhaus: I have pushed https://github.com/clearhaus/gateway-api-docs/pull/92/commits/b20346aa7f0a5dd3aa4887224d9fd8ad485b10f0.

mt-clearhaus:
Deployed.

PR #89, opened 2019-08-21
Require signing

ct-clearhaus:
- [ ] Changelog

PR #93, opened 2020-01-09
CSC is optional only when partner is trusted

ct-clearhaus:
Cf. https://github.com/clearhaus/gateway-api-docs/pull/89#issuecomment-572486179


mt-clearhaus:
Deployed :heavy_check_mark:

PR #95, opened 2020-02-19
csc matches may be null

mt-clearhaus:
This is one possible way to handle it.

I currently lean more towards this than to add a superfluous `checked` attribute. Even with such an attribute `matches` must still be allowed to be `null` or absent.

### DIFF
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -190,6 +190,12 @@ match; in this case the <code>status</code> <code>code</code> will be
 authorizations may be in place for a merchant.
 </p>
 
+<p class="alert alert-info">
+<b>Notice:</b> <code>csc</code> <code>matches</code> may be <code>null</code>
+if the card scheme or the issuer decided not to check the CSC or if the card
+scheme failed checking the CSC on behalf of the issuer.
+</p>
+
 
 ### Withdraw money
 
@@ -867,6 +873,13 @@ https://gateway.clearhaus.com/account
 Follow coming changes on the [source code repository](https://github.com/clearhaus/gateway-api-docs).
 
 Sorted by descending timestamp.
+
+### Attribute `csc matches` in a response may be `null`
+
+Starting 2020-xx-xx, `csc matches` in a response to a `POST /authorizations`
+request may be `null`. The value covers the case where the card scheme or the
+issuer decided not to check the CSC and the case where the card scheme failed
+checking the CSC on behalf of the issuer.
 
 ### Request signing becomes mandatory
 


dvaeversted:
I think this would be a good solution, would allow us to differentiate between if the check failed or simply not performed.
Also i think it would be preferred over adding yet another field about the CSC.

ct-clearhaus:
> allow us to differentiate between if the check failed or simply not performed.

How does it affect the decision you take regarding the transaction processing?

dvaeversted:
> How does it affect the decision you take regarding the transaction processing?

If the CSC is not valid, we would prefer to simply drop the transaction since it is an indicator of the cardholder not having the correct card details. Our merchants would be much surprised to learn in connection with a chargeback that the CSC was not valid, but the transaction was still approved. 

This however raises the concern, that if the issuer did not validate the CSC at all based upon other factors, the expected result for the merchant would be that the transaction is allowed to continue.

ct-clearhaus:
Perhaps I'm dumb, but I don't get it.

Let me flesh out all possible situations and the decision I would take:

| checked | confirmed_to_match | 3DS   | Decision |
| ------- | ------------------ | ----- | -------- |
| true    | true               | true  | allow capture |
| true    | false              | true  | allow capture |
| false   | false              | true  | allow capture |
| true    | true               | false | allow capture |
| true    | false              | false | be careful; consider risk |
| false   | false              | false | be careful; consider risk |

Thus, my decision is "allow capture" if and only if `3DS || (checked && confirmed_to_match)` is `true`; if `false`, then my decision is "be careful; consider risk".

In other words, it is sufficient to receive `checked && confirmed_to_match` as we currently do with `csc.matches` being a boolean.

Is your decision matrix different? If so, how is it different, and why?

dvaeversted:
`checked && !confirmed_to_match && !3DS`
We would like to decline this transaction, since it signals that wrong data was entered.

`!checked && !confirmed_to_match && !3DS`
We would probably require it to be stepped up to 3DS, to mitigate the risk.


It is not realistic to inform merchants that a transaction is of higher risk, we have had that possibility before and it ends up with the merchant being angry over the fact that such a transaction was accepted at all.

ct-clearhaus:
So mine "be careful; consider risk" you'd translate to "disallow capture; attempt to auto-void", but the decision logic is still `3DS || (checked && confirmed_to_match)`, right?

dvaeversted:
Yes

ct-clearhaus:
In conclusion, the current `matches` which is [logical equivalent](https://en.wikipedia.org/wiki/Logical_equivalence) to `checked && confirmed_to_match` is sufficient, right?

dvaeversted:
The only case is the one where we have provided a CSC, but the following logic is true `!checked && !confirmed_to_match && !3DS` and the authorization is approved.

I do not know if that can ever happen in the wild, but right now we would consider that as a decline and void the transaction.
So if that is never a case we should consider, then the current data provided would indeed suffice.

ct-clearhaus:
> The only case is the one where we have provided a CSC, but the following logic is true `!checked && !confirmed_to_match && !3DS` and the authorization is approved.
> 
> I do not know if that can ever happen in the wild, but right now we would consider that as a decline and void the transaction.

It wouldn't be surprised to see it happen. (🤔 It could be an issuer that just approves without CSC check for "low enough amount".) I included the situation in the table:

<kbd>
Screenshot 2020-02-20 at 12 46 34
</kbd>

So the logic
```
if 3DS || (checked && confirmed_to_match) then
  approved
else
  do not allow capture
  attempt to void
end-if
```
will work. And the condition `3DS || (checked && confirmed_to_match) === 3DS || matches`, so the current `matches` is still sufficient, right?

ct-clearhaus:
In the real world, I would probably add a condition for the amount; that is, if the amount is "low enough", then as a merchant, I would likely consider an authorisation where `!(3DS || matches)` as approved—simply because the probability of receiving a chargeback is neglegible if the amount is "low enough".

Now, what this "low enough" threshold would actually be quite likely depends on several factors; among others merchant country, card country, market, and risk appetite.

Add in that chargebacks is not everything; merchants can also face issues with just fraud markings ("TC40" and "SAFE" for Visa and Mastercard, respectively).

mt-clearhaus:
I dare to say that it appears https://github.com/clearhaus/gateway-api-docs/pull/96 wins over this PR. I will postpone closing the PR a few days, though.

kse-clearhaus:
PR #96 has been merged into documentation. Will probably be live in a couple of minutes.

PR #97, opened 2020-04-15
Reformatting

kse-clearhaus:
This is the initial samsung pay api description.

I've decided to change the DL to vertical, to avoid truncating the input fields.
This should also be done to other `dl`'s before this is to be merged.


source/index.html.md
```diff
@@ -630,6 +630,31 @@ object][ApplePay-PaymentToken] for more information.
 </p>
 
 
+##### Method: `samsungpay`
+
+Samsung Pay provides the payment details as either JWE or JWE/JWS JOSE objects. Currenlty we only
+accept the JWE.
+Your system must extract the content encryption key (<code>cek</code>) used to encrypt the JWE
```
ct-clearhaus, line 637:
```suggestion
Your system must extract the content encryption key (CEK) used to encrypt the JWE
```

... but we don't seem to use the CEK acronym anywhere?

source/index.html.md
```diff
@@ -630,6 +630,31 @@ object][ApplePay-PaymentToken] for more information.
...
+payload and Base64-encode it, along with the entire JWE.
+
+<dl class="dl-vertical">
+  <dt>samsungpay[jwe]</dt>
```
ct-clearhaus, line 641:
```suggestion
  <dt>samsungpay[token]</dt>
```

... or something alike? Granted, it's JWE, but we could also use `samsungpay[json]` which I also dislike. What's the **name** of that fellah in Samsung Pay lingo?

source/index.html.md
```diff
@@ -630,6 +630,31 @@ object][ApplePay-PaymentToken] for more information.
...
+  <dd>
+    Type: <a href="https://tools.ietf.org/html/rfc7516#section-3.1">JWE Compact serialization</a>
+    <br>
+    <i>Required</i>
+    <br>
+    Complete payment token as received from Samsung Pay.
+  </dd>
+  <dt>samsungpay[symmetric_key]</dt>
```
ct-clearhaus, line 649:
```suggestion
  <dt>samsungpay[cek]</dt>
```
... permaybehaps? Then the description below could introduce the "CEK" acronym (and we could shorten the introductory text)?

source/index.html.md
```diff
@@ -630,6 +630,31 @@ object][ApplePay-PaymentToken] for more information.
...
+  <dd>
+    Type: Base-64 encoded string
```
ct-clearhaus, line 651: Mention format to be something like `[:base64:]{16}`?

source/index.html.md
```diff
@@ -630,6 +630,31 @@ object][ApplePay-PaymentToken] for more information.
 </p>
 
 
+##### Method: `samsungpay`
+
+Samsung Pay provides the payment details as either JWE or JWE/JWS JOSE objects. Currenlty we only
```
ct-clearhaus, line 635:
```suggestion
Samsung Pay provides the payment details as either JWE or JWE/JWS JOSE objects. Currently we only
```

source/index.html.md
ct-clearhaus: Would we like to receive "the JWE/JWS" in another parameter? What would the name of that one be?

source/index.html.md
```diff
@@ -630,6 +630,31 @@ object][ApplePay-PaymentToken] for more information.
...
+accept the JWE.
+Your system must extract the content encryption key (<code>cek</code>) used to encrypt the JWE
+payload and Base64-encode it, along with the entire JWE.
+
+<dl class="dl-vertical">
+  <dt>samsungpay[jwe]</dt>
+  <dd>
+    Type: <a href="https://tools.ietf.org/html/rfc7516#section-3.1">JWE Compact serialization</a>
+    <br>
+    <i>Required</i>
```
ct-clearhaus, line 645: Elsewhere we mark optional parameters since the majority of parameters are required 🙂

source/index.html.md
kse-clearhaus: This is to refer to the JWE spec: https://tools.ietf.org/html/rfc7516#section-2

It should be upper case however.

source/index.html.md
kse-clearhaus: It's not JSON. I think `jwt` [JSON web token](https://en.wikipedia.org/wiki/JSON_Web_Token) could be the correct lingo.

source/index.html.md
kse-clearhaus: It could be changed as you suggested, but I'd rather not remove the reference from the introduction text. I want to be clear about what has to happen. Users should not just encounter `cek` and then have to look up where that comes from. I want to be explicit.

source/index.html.md
kse-clearhaus: But then there's overloading, since JWT can both refer to JWS and JWE.

source/index.html.md
kse-clearhaus: Well, the Base64-string will be 28 characters long. But that's not enough, because 19, 20 and 21 bytes Base64-encoded are all 28 characters.

I'm experimenting with a less obtuse pseudo-regexp syntax `[:...:]`.

I'll give it some thought.

kse-clearhaus: I'll save that for another day, `[:base64:]` it is.

mt-clearhaus: I saw "Base64-encode" earlier in this PR and notice that "Base-64 encoded" is inconsistent with that.

source/index.html.md
```diff
@@ -630,6 +630,31 @@ object][ApplePay-PaymentToken] for more information.
...
+    <br>
+    Complete payment token as received from Samsung Pay.
+  </dd>
+  <dt>samsungpay[symmetric_key]</dt>
+  <dd>
+    Type: Base-64 encoded string
+    <br/>
+    <i>Required</i>
+    <br/>
+    16 byte symmetric AES key used to encrypt the payment token, Base64 encoded.
```
mt-clearhaus, line 655: "Base64 encoded" is a third version :slightly_smiling_face:

source/stylesheets/all.css.scss
```diff
@@ -124,3 +124,12 @@ pre code.highlight {
     display: none;
   }
 }
+
+dl.dl-vertical {
+  margin-left: 10px;
+}
+
+/* On vertial description list, indent the description slightly to enhance */
```
mt-clearhaus, line 132:
```suggestion
/* On vertical description list, indent the description slightly to enhance */
```

source/index.html.md
kse-clearhaus: 3 different way on 34 lines, that's almost impressive :D

source/index.html.md
ct-clearhaus: I think it is obvious that we do not disagree that it should be clear what has to happen. Though, I suspect we may disagree that the only way to achieve this is to introduce "CEK" in the text before the parameter 🙂

I think it's not much different than encountering `currency` and then in the next line notice that it's the alpha-3 version 🙂

However, I am certainly [not sure about the naming](https://twitter.com/secretGeek/status/7269997868) — if `symmetric_key`, or `cek` or something else is the better name.

kse-clearhaus: I've changed it around a bit, forgot to push the changes :)

Please give it a look now.

source/index.html.md
ct-clearhaus: These 16 bytes will always end up as a 24 char long string when Base64 encoded, right?

If I am not mistaken, let's be stringent and write `[:base64:]{24}` 🙂

source/index.html.md
```diff
@@ -630,6 +630,27 @@ object][ApplePay-PaymentToken] for more information.
...
+  <dd>
+    <a href="https://tools.ietf.org/html/rfc7516#section-3.1"><span class="type">Compact serialized JWE token</span></a>
+    <br>
+    Complete JWE token as received from Samsung Pay.
+  </dd>
+  <dt>samsungpay[cek]</dt>
+  <dd>
+    <span class="type">[:base64:]</span>
+    <br/>
+    16 byte symmetric AES content encryption key (CEK) used to encrypt the
```
ct-clearhaus, line 649:
```suggestion
    The (symmetric AES) Content Encryption Key (CEK) used to encrypt the
```
I guess nobody would need to know about the number of bytes? I suspect it's only relevant to document the length of the Base64 encoded key (which is always 24 if I am not mistaken)?

source/index.html.md
```diff
@@ -630,6 +630,27 @@ object][ApplePay-PaymentToken] for more information.
...
+Currently only the JWE is accepted.
+
+<dl class="dl-vertical">
+  <dt>samsungpay[jwt]</dt>
+  <dd>
+    <a href="https://tools.ietf.org/html/rfc7516#section-3.1"><span class="type">Compact serialized JWE token</span></a>
+    <br>
+    Complete JWE token as received from Samsung Pay.
+  </dd>
+  <dt>samsungpay[cek]</dt>
```
ct-clearhaus, line 645: If we eventually support JWE/JWS, will it still make sense to accept `cek` as the parameter?

source/index.html.md
```diff
@@ -630,6 +630,27 @@ object][ApplePay-PaymentToken] for more information.
...
 
+##### Method: `samsungpay`
+
+Samsung Pay provides the payment details as either JWE or JWE/JWS JOSE objects.
+Currently only the JWE is accepted.
+
+<dl class="dl-vertical">
+  <dt>samsungpay[jwt]</dt>
+  <dd>
+    <a href="https://tools.ietf.org/html/rfc7516#section-3.1"><span class="type">Compact serialized JWE token</span></a>
```
ct-clearhaus, line 641: I am no HTML expert, but wouldn't it normally be `<span><a>text</a></span>`?

ad-clearhaus:
As the only thing not commented on: I like the styling changes. :smile:  :+1:

source/index.html.md
kse-clearhaus: Well, I chose to write 16 bytes because the algorithm is 128bit AES, this was just to emphasize what key size we're talking about. I'm trying to use the text to give hints about what to expect.

Though I do think it should be changed to 128 bits.

source/index.html.md
kse-clearhaus: Yes. Unless we ask for their private key.

source/index.html.md
kse-clearhaus: For `span` it doesn't matter. 

But I've changed it to `div` in my local code, it matters there. So I've swapped places for them.

source/index.html.md
kse-clearhaus: Sure.

source/index.html.md
```diff
@@ -630,6 +630,27 @@ object][ApplePay-PaymentToken] for more information.
 </p>
 
 
+##### Method: `samsungpay`
+
+Samsung Pay provides the payment details as either JWE or JWE/JWS JOSE objects.
+Currently only the JWE is accepted.
+
+<dl class="dl-vertical">
+  <dt>samsungpay[jwt]</dt>
```
ct-clearhaus, line 735: I can't help 👃 the name ... sorry 🙏 ... that is, I still dislike that we don't have a *name* for the parameter but us the "format" as the parameter name.

Would we want to accept "JWE tokens" and "JWE/JWS tokens" in the same parameter when/if we support both? With `[3dsecure][v2]` we ask to get `ares` and `rreq` in two different parameters — perhaps this was a design flaw?

In Samsung Pay - Web checkout - Integration guide, the name "payment credential data" and just "credential data" is used. Perhaps `samsungpay[credential_data]` would make sense?

kse-clearhaus: You don't have to be sorry, we're being pedantic about docs, that's fine :)

Let me just state my preferences:

- If we decide to accept both JWE and JWS as the same parameter, I prefer `jwt`
- If we decide to accept the separately, I prefer `jwe` and `jws`.

It's common around the web to use JWE and JWS as nouns. We could also use `jwt_string`, but that's uglier to me. (The guide you referred to and the other guides also use JWE as a noun)

When I read `samsungpay[jwt]` I read it as _"the Samsung pay JWT"_, which makes perfect sense, as the JWT is in a context here.


However _"the Samsung pay credential data"_ doesn't immediately make sense, is it the unpacked JWT payload? The only thing it has going for it is that it's used in the web checkout guide.

But I don't want to bikeshed this anymore, you're free to select whichever you like :)


Now, if we choose to also accept the JWS, then we should likely follow the 3DSv2 precedence and split them to accept either the JWS or the JWT.
(Though admittedly, I would strongly prefer them to be a single parameter)

kse-clearhaus:
I updated the gems to fix some vulnerabilities btw.

source/index.html.md
```diff
@@ -630,6 +770,35 @@ object][ApplePay-PaymentToken] for more information.
...
+  <dt>
+    samsungpay[jwt]
+    <span class="type">
+      <a href="https://tools.ietf.org/html/rfc7516#section-3.1">
+        Compact serialized JWE token
+      </a>
+    </span>
+  </dt>
+  <dd>
+    Complete JWE token as received from Samsung Pay.
```
ct-clearhaus, line 788: Perhaps include an example snippet? (Cf. https://github.com/clearhaus/gateway-api-docs/pull/67)

source/index.html.md
```diff
@@ -630,6 +770,35 @@ object][ApplePay-PaymentToken] for more information.
...
+  <dt>
+    samsungpay[cek]
+    <span class="type">
+      [:base64:]{24}
+    </span>
+  </dt>
+  <dd>
+    The 128 bit symmetric AES content encryption key (CEK) used to encrypt the
+    payment token, Base64 encoded. This CEK must be extracted from the JWE
+    token.
```
ct-clearhaus, line 799: `</dd>` missing

source/index.html.md
```diff
@@ -630,6 +770,35 @@ object][ApplePay-PaymentToken] for more information.
...
+    Complete JWE token as received from Samsung Pay.
+  </dd>
+  <dt>
+    samsungpay[cek]
+    <span class="type">
+      [:base64:]{24}
+    </span>
+  </dt>
+  <dd>
+    The 128 bit symmetric AES content encryption key (CEK) used to encrypt the
```
ct-clearhaus, line 797: Tried to find mention of the name "content encryption key" in Samsung Pay documentation 🤞 , but 😞
The Web Checkout Integration Guide mentions `plainEncKey` but 🚫

Perhaps `symmetric_key` would be as good as `cek`? And then at least it'd align with https://github.com/clearhaus/gateway-api-docs/pull/97/files#diff-d30d91dbcd59693b69f9242acb4630c5R683 🙁

source/index.html.md
ct-clearhaus: W3 isn't too happy about the example though; see 2nd error of https://validator.w3.org/nu/?doc=https%3A%2F%2Fdocs.gateway.clearhaus.com%2F 🙂

source/index.html.md
kse-clearhaus: Nicely spotted :) I assume you ran it through the validator?

source/index.html.md
kse-clearhaus: Good idea to add examples :+1: 

I was planning to, but honestly I was considering finding some JS/CSS-thing that could help us with collapsible examples.

source/index.html.md
kse-clearhaus: So, I agree that `symmetric_key` is accurate. But since this is a JWE, I'd like to be explicit about what exactly this is in the terms of a JWE. Here it is very explicit, since a JWE could technically also contain a second symmetric key.

(If you ask me, the Samsung guides use bad naming)

source/index.html.md
ct-clearhaus: I just stumbled upon it randomly and that reminded me to check a validator (but didn't dare to run it on the PR version; couldn't immediately figure out an easy way for the online W3 thingy) 🤷‍♂️

kse-clearhaus:
@clearhaus/pci: I've added my initial thoughts on how the JWS API should be.

WDYT?

You're welcome to ignore naming for now, we can discuss that later.
I've just used the naming from the Samsung guide.

2020-05-11-121646_983x791_scrot

ct-clearhaus:
May I suggest that we split this into two PRs "reformat the hell out of the API docs" and "add samsung pay"? 😬

I'd be fine with mutating this PR into the former have it merged quickly 🙂

Thus, I'll comment only on the reformatting here.

Although I dislike the "regexp-y" description of the format being "hidden" to the very right (I fear it'll not attract the attention it deserves), it should be shoved to the right consistently:

<kbd>
Screenshot 2020-05-11 at 14 32 07
</kbd>

Maybe some "right inner margin" adjustment?
<kbd>
Screenshot 2020-05-11 at 14 41 08
</kbd>

mt-clearhaus:
I believe `3Dsecure.io` should be changed to `3dsecure.io` or `3DSecure.io`.

davidfromdk:
Agree with splitting it up into two PRs :-) As a quick fix one could steal the spacing/margin properties used here: https://getbootstrap.com/docs/4.3/components/card/#header-and-footer

ct-clearhaus:
As for the Samsung Pay part:
* Shouldn't we just include it directly in the docs? (Both for test and production?)
* "`card`, `applepay`, `mobilepayonline`." ->"`card`, `applepay`, `mobilepayonline`, `samsungpay`."
* "Only the JWS is accepted" -> "Only the JWE/JWS is accepted"?

kse-clearhaus:
> As for the Samsung Pay part:
> 
> * Shouldn't we just include the public key directly in the docs? (Both for test and production?)
> * "`card`, `applepay`, `mobilepayonline`." ->"`card`, `applepay`, `mobilepayonline`, `samsungpay`."
> * "Only the JWS is accepted" -> "Only the JWE/JWS is accepted"?

Yeah, I agree that we could consider that.

It was more an attempt at doing nudging, so their initial thought is that they have to decrypt it.
But it's not sensitive, no.

I'll add it :+1:

ct-clearhaus:
> It was more an attempt at doing nudging, so their initial thought is that they have to decrypt it.

That's a really good point. I guess there's already incentive for the client to have their own keypair and derive the CEK — namely to support recurring payments. Perhaps we should leave this piece for @clearhaus/partner to decide (and leave it for them to hand out the public key also)?

ct-clearhaus:
I'm making this PR "reformat only"; however, I'll not rename the branch — because then GitHub will close the PR.

kse-clearhaus:
Not sure about the left border. WDYT?

2020-05-12-153438_735x902_scrot

kse-clearhaus:
Maybe just this:

2020-05-12-154936_995x917_scrot

davidfromdk:
I prefer the one without the left border. Maybe move the `POST https://gateway.clearhaus.com/credits` above the "Parameters" heading?

ct-clearhaus:
> Maybe move the `POST https://gateway.clearhaus.com/credits` above the "Parameters" heading?

Arh, that's a funny thing. It's like this (somewhere) because for credits there are different parameters for the two different endpoints:
<kbd>
Screenshot 2020-05-12 at 22 09 01
</kbd>

But doubly funny, that deprecation is actually removed in production, so the documentation should go as well. And then it should be fixed for refunds and voids also.

But sadly, it's actually removed on `master`, so I suspect @kse-clearhaus branched out a little too early — on `8446100dadd02f637332ef0536c0cb3f36efbcd7` from July 2019 to be precise 😢

@kse-clearhaus, you wanna handle the merge of master into this PR to bring it up to date? I'll allow myself to take my hands off since I'll probably just screw it up 🙂

kse-clearhaus:
Hah, I have no idea how that happened. I'll handle the merge, of course :+1:

kse-clearhaus:
2020-05-14-142759_805x1163_scrot

kse-clearhaus:
I'm going to grab the previous approval and do a deployment :)

kse-clearhaus:
BTW, my guess/hope is that we're not using Heroku anymore:

```
$ dig +short docs.gateway.clearhaus.com
d2te2qckkr0lw4.cloudfront.net.
143.204.237.66
143.204.237.35
143.204.237.83
143.204.237.7
```

So deployment should happen automatically on merge to master :tada:

PR #106, opened 2020-06-25
Add EMV Token by expanding payment methods

ad-clearhaus:
Add EMV Token by expanding payment methods? Discuss.


mt-clearhaus:
We shall of course remember to add `samsungpay` later, if we choose to go in this direction.

ad-clearhaus:
Yes of course. I have sporadically mentioned samsungpay/googlepay. But they are not included on `master` yet, so left them out of the actual text.

PR #107, opened 2020-06-26
Add EMV Token by expanding card method

ad-clearhaus:
Add EMV Token by expanding card method? Discuss.

### DIFF
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -565,6 +565,24 @@ Exactly one payment method must be used.
     Card Security Code.
     <div class="type">Optional when partner is trusted</div>
   </dd>
+  <dt>card[tav]
+    <span class="type">[:base64:]{19,21}</span>
+  </dt>
+  <dd>
+    Token Authentication Value
+    <div class="type">Must only be set on token payments</div>
+  </dd>
+  <dt>card[token_payment_method_indication]
+    <span class="type">[\x20-\x7E]{7,15}
+    <a target="_blank" href="http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">
+      ASCII printable characters
+     </a>
+    </span>
+  </dt>
+  <dd>
+    Indicate which token payment method used. Either: <code>applepay</code>, <code>samsungpay</code>, <code>mobilepayonline</code>
+    <div class="type">Can only be set on token payments</div>
+  </dd>
   <dt>card[pares]
     <span class="type">[:base64:]</span>
   </dt>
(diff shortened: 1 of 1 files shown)


source/index.html.md
```diff
@@ -565,6 +565,24 @@ Exactly one payment method must be used.
...
+  <dt>card[token_payment_method_indication]
+    <span class="type">[\x20-\x7E]{7,15}
+    <a target="_blank" href="http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">
+      ASCII printable characters
+     </a>
+    </span>
+  </dt>
+  <dd>
+    Indicate which token payment method used. Either: <code>applepay</code>, <code>samsungpay</code>, <code>mobilepayonline</code>
+    <div class="type">Can only be set on token payments</div>
```
mt-clearhaus, line 584:
```suggestion
    Indicate token payment method used. We currently support: <code>applepay</code>, <code>samsungpay</code>, <code>mobilepayonline</code>
    <div class="type">Must only be set on token payments</div>
```

To be similar to line 573.

source/index.html.md
```diff
@@ -583,6 +601,13 @@ Exactly one payment method must be used.
   <br />
   <b>Notice:</b> An authorization that includes <code>card[pares]</code> and/or
   <code>card[csc]</code> cannot be a subsequent recurring authorization.
+  <br />
+  <b>Notice:</b> If the payment was made through a token based payment method, the token authentication value must be included
```
mt-clearhaus, line 605:
```suggestion
  <b>Notice:</b> If the payment was made through a token based payment method, the token authentication value (<code>card[tav]</code>) must be included
```

Or just something similar :slightly_smiling_face:

source/index.html.md
```diff
@@ -583,6 +601,13 @@ Exactly one payment method must be used.
...
+  <br />
+  <b>Notice:</b> The <code>token_payment_method_indication</code> is used for
+  payment method categorization which influences how the transactions is
```
mt-clearhaus, line 608:
```suggestion
  payment method categorization which influences how the transactions are
```

source/index.html.md
ad-clearhaus: The must vs. can was deliberate. If `token_payment_method_indication` is unknown, it may be left out. A `tav` may never be left out.
And know, come to think of it, on subsequent recurring, we are missing a `tav`. But then again, should subsequent recurring just be completed as a regular card payment?

ad-clearhaus: I am adding a note in https://github.com/clearhaus/issues-pci/issues/1604

mt-clearhaus: "Must" doesn't imply that the parameter must always be included. It only implies that it must not be included for non-token payments. But I can live with "can" :slightly_smiling_face:

PR #81, opened 2019-08-14
Introduce series[] using a reference

ct-clearhaus:
Supports installments (`installment`), recurring (`recurring`), UCOF (`unscheduled`) and CIT-only COF (`cardholder`) via `series[type]` and requires a reference to connect transactions in the series via `series[previous]`.

Warning! The naming is only a suggestion. Perhaps we can find better names?

PR #79, opened 2019-07-11
Add series[] and deprecate recurring

ct-clearhaus:

PR #103, opened 2020-05-15
Add samsungpay (again)

ct-clearhaus:
Ref. #97, #100, and #102 🙂

PR #111, opened 2020-09-29
Introduce credential, initiator and series[]

ct-clearhaus:
* `credential` to indicate if stored credential is being stored or used (or
    irrelevant, by leaving it out)
* `initiator` to indicate what party initiate the authorization
* `series[]` to replace `recurring` but also extend with UCOF and
  Clearhaus-internal and raw transaction reference to the previous-in-series

This PR supersedes:
* https://github.com/clearhaus/gateway-api-docs/pull/78
* https://github.com/clearhaus/gateway-api-docs/pull/79
* https://github.com/clearhaus/gateway-api-docs/pull/81
* https://github.com/clearhaus/gateway-api-docs/pull/83

## To do

* [x] Clean-up other explanations
* [x] Reference partner guideline, and potentially move some text over there

### DIFF
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -591,11 +591,13 @@ Exactly one payment method must be used.
     cardholder if the cardholder decided the transaction should be created.
     This is regardless of whether stored payment credential is being used.
     <br />
-    For compliance reasons there should be a previous approved transaction
-    marked with <code>storing</code> before <code>initiator</code> may be
-    <code>merchant</code>.
-    For subsequent-in-series authorizations, <code>merchant</code> is assumed;
-    otherwise, <code>cardholder</code> is assumed.
+    For compliance reasons there should be a previous approved transaction (for
+    the cardholder and the merchant) marked with <code>storing</code> before
+    <code>initiator</code> may be <code>merchant</code>.
+
+    For subsequent-in-series authorizations, <code>cardholder</code> is invalid
+    and <code>merchant</code> is assumed; otherwise, <code>cardholder</code> is
+    assumed.
 
     <div class="type">Optional</div>
   </dd>
@@ -617,7 +619,7 @@ Exactly one payment method must be used.
     <br />
     <code>recurring</code>: A series of transactions where the cardholder has
     explicitly agreed that the merchant will repeatedly charge the cardholder at
-    regular, predetermined intervals.
+    regular, predetermined intervals that may not exceed 1 year.
     <br />
     <code>unscheduled</code>: A series of transactions where the cardholder has
     explicitly agreed that the merchant will repeatedly charge the cardholder at
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -682,8 +682,19 @@ Exactly one payment method must be used.
   </dt>
   <dd>
     Deprecated! Please use <code>series</code>.
+    <br />
+    <div class="type">Optional. Cannot be used with <code>series</code> or
+    <code>initiator</code>.</div>
   </dd>
 </dl>
+
+<p class="alert alert-info">
+  <b>Notice:</b> When <code>recurring</code> is used, Clearhaus automatically
+  identify if there was a previous-in-series and together with the level of
+  authentication (CSC, 3-D Secure, etc.) conclude if the payment is a first
+  recurring or a subsequent recurring.
+</p>
+
 
 ##### Method: `card`
 
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -378,7 +378,7 @@ Example response (snippet):
 This should be followed by a capture.
 
 Subsequent recurring authorizations initiated by the merchant are made
-similarly, however, CSC would not be included, and the previous-in-series is
+similarly, however, CSC is not included, and the previous-in-series is
 referenced, e.g.:
 
 ````shell
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -927,7 +927,7 @@ The Mastercard specific reference to the series contains the following parts.
     Trace ID being the concatenation of values
     Data element 63 subfield 1 (Financial Network Code) (position 1-3),
     Data element 63 subfield 2 (Banknet Reference Number) (position 4-9),
-    Data element 15 (Date, Settlement) (position 10-13), and
+    Data element 15 (Date, Settlement, in MMDD format) (position 10-13), and
     two spaces;
     to be used in Data Element 48, Subfield 63.
 
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -1402,27 +1402,6 @@ category, the name might be necessary for approval.
 
 In the first quarter of 2020 signing of POST requests will become mandatory. We will work together with clients to ensure their requests are compliant before introducing enforcement of the requirement in the transaction gateway.
 
-### Remove deprecated `/cards`, `threed_secure` and `card[number]`
-
-The deprecations announced in 2018 will be executed. Partners must expect
-`/cards` endpoints, the `threed_secure` and the `card[number]` parameters to be
-unavailable at any point in time after 2019-08-20.
-Refer to [the documentation source code
-changes](https://github.com/clearhaus/gateway-api-docs/pull/82/files) for the
-exact documentation change.
-
-### Add merchant blocked by cardholder status
-
-Starting 2019-04-11 a new status code `40420 - Merchant blocked by cardholder`
-is avaliable. Please be advised that appropriate action must be taken to
-adequately handle this status code, see [Merchant blocked by cardholder
-status](#merchant-blocked-by-cardholder-status)
-
-### Add a valid test CSC
-
-Starting 2019-02-22 the CSC 987 has been added as a valid CSC for all PANs when
-testing against `gateway.test.clearhaus.com`.
-
 [JSON-HAL]: https://tools.ietf.org/html/draft-kelly-json-hal-08 "IETF HAL draft"
 [HATEOAS]: http://en.wikipedia.org/wiki/HATEOAS
 [Tokenization]: http://en.wikipedia.org/wiki/Tokenization_(data_security)
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -1141,6 +1141,11 @@ POST https://gateway.clearhaus.com/credits
   </dd>
 </dl>
 
+<p class="alert alert-info">
+<b>Notice:</b> Implicitly, <code>initiator</code> is <code>merchant</code> and
+<code>credential</code> is <code>used</code>.
+</p>
+
 ### Account
 
 The account resource holds basic merchant account information. Only `HTTP GET`
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -932,6 +932,33 @@ The Mastercard specific reference to the series contains the following parts.
   </dd>
 
   <dt>
+    series[previous][mastercard][exemption]
+    <span class="type">(<code>mit</code>|<code>recurring</code>)</span>
+  </dt>
+  <dd>
+    The <i>Mastercard exemption</i> to be used for the series. If the
+    previous-in-series is a subsequent-in-series it should be equal to the
+    <i>Mastercard exemption</i> applied for the previous-in-series.
+    <ul>
+      <li><code>mit</code>: A <i>Mastercard exemption</i> (not formally an
+      acquirer exemption for SCA) indicating that the transaction is out of
+      scope for SCA. Relevant for varying-amount series. Refer to the partner
+      guideline for more details.</li>
+      <li><code>recurring</code>: The previous-in-series was acquirer exempted
+      for SCA because it was a subsequent-in-series, fixed-amount series
+      authorization.</li>
+    </ul>
+    Clearhaus will use the same <i>Mastercard exemption</i> for the entire
+    series when later referring to the previous-in-series via
+    <code>series[previous][id]</code>.
+
+    <div class="type">Conditional.
+    Required if <code>series[previous][mastercard]</code> is present.
+    Cannot be present if <code>series[previous][id]</code> or
+    <code>series[previous][visa]</code> is present.</div>
+  </dd>
+
+  <dt>
     series[previous][mastercard][tokenized]
     <span class="type">(true|false)</span>
   </dt>
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -394,8 +394,8 @@ curl -X POST \
   -H "Signature: <signing-api-key> RS256-hex <signature>"
 ````
 
-A first-in-series authorization can also be made using the `applepay` or
-`mobilepayonline` payment methods; subsequent-in-series authorizations, however,
+A first-in-series authorization can also be made using the `applepay`
+payment method; subsequent-in-series authorizations, however,
 must be made using the `card` payment method using the card details of the
 previous-in-series authorization referenced.
 
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -925,7 +925,10 @@ The Mastercard specific reference to the series contains the following parts.
     two spaces;
     to be used in Data Element 48, Subfield 63.
 
-    <div class="type">Optional. Cannot be present if <code>series[previous][id]</code> or <code>series[previous][visa]</code> is present.</div>
+    <div class="type">Conditional.
+    Required if <code>series[previous][mastercard]</code> is present.
+    Cannot be present if <code>series[previous][id]</code> or
+    <code>series[previous][visa]</code> is present.</div>
   </dd>
 
   <dt>
@@ -935,7 +938,10 @@ The Mastercard specific reference to the series contains the following parts.
   <dd>
     Indicate if the series was initiated with a tokenized payment.
 
-    <div class="type">Conditional. Required if <code>series[previous][mastercard][tid]</code> is present.</div>
+    <div class="type">Conditional.
+    Required if <code>series[previous][mastercard]</code> is present.
+    Cannot be present if <code>series[previous][id]</code> or
+    <code>series[previous][visa]</code> is present.</div>
   </dd>
 </dl>
 
@@ -950,7 +956,9 @@ The Visa specific reference to the series has only one part.
     Transaction ID from Field 62.2 of the first (or previous) in series
     authorization; to be used in Field 125, Usage 2, Dataset ID 03.
 
-    <div class="type">Optional. Cannot be present if <code>series[previous][id]</code> or <code>series[previous][mastercard]</code> is present.</div>
+    <div class="type">Optional.
+    Cannot be present if <code>series[previous][id]</code> or
+    <code>series[previous][mastercard]</code> is present.</div>
   </dd>
 </dl>
 
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -567,13 +567,13 @@ Exactly one payment method must be used.
   </dt>
   <dd>
     Indicate if the payment credential (e.g. PAN and expiry) will be stored for
-    a later transaction where the payment credential is not provided by the
+    future use where the payment credential is not provided by the
     cardholder but collected from (encrypted) storage.
     <br />
     <code>storing</code>: The payment credential will be stored; may only be
     stored if the authorization is approved.
     <br />
-    <code>using</code>: The payment credential was already stored and are now
+    <code>using</code>: The payment credential has already been stored and is now
     being used.
     <br />
     If <code>initiator</code> is <code>merchant</code> then
@@ -651,8 +651,6 @@ Exactly one payment method must be used.
 
     <div class="type">Conditional. Cannot be present if <code>series[type]</code> is present.</div>
   </dd>
-
-
   <dt>text_on_statement
     <span class="type">[\x20-\x7E]{2,22}
       <a target="_blank" href="http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">
@@ -904,7 +902,7 @@ Only one 3-D Secure version can be used for a given authorization.
 
 If the previous-in-series authorization was made via this API, you should use
 `series[previous][id]` to reference the previous-in-series authorization.
-Otherwise, you must obtain explicit approval from Clearhaus to use these raw
+Otherwise, you must obtain explicit approval from Clearhaus to use the raw
 scheme values grouped in <code>series[previous][mastercard]</code> and
 <code>series[previous][visa]</code>.
 
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -937,7 +937,7 @@ The Mastercard specific reference to the series contains the following parts.
     <span class="type">(true|false)</span>
   </dt>
   <dd>
-    Indicate if the series was initiated with a tokenized payment method.
+    Indicate if the series was initiated with a tokenized payment.
 
     <div class="type">Conditional. Required if <code>series[previous][mastercard][tid]</code> is present.</div>
   </dd>
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -367,19 +367,6 @@ curl -X POST \
   -H "Signature: <signing-api-key> RS256-hex <signature>"
 ````
 
-Example response (snippet):
-
-````json
-{
-    "id": "e3e9d215-6efc-4c0e-b3d7-2226057c6de8",
-    "status": {
-        "code": 20000
-    },
-    "processed_at": "2018-07-09T12:58:56+00:00",
-    "recurring": true
-}
-````
-
 This should be followed by a capture except when the amount is `0`.
 
 Subsequent authorizations are made similarly, however, CSC would not be
@@ -540,12 +527,52 @@ Exactly one payment method must be used.
   <dd>
     Amount in minor units of given currency (e.g. cents if in Euro).
   </dd>
+
   <dt>currency
     <span class="type">[A-Z]{3}</span>
   </dt>
   <dd>
     <a target="_blank" href="currencies.txt">3-letter currency code</a>. (Some exponents differ from ISO 4217.)
   </dd>
+
+  <dt>credential
+    <span class="type">(<code>storing</code>|<code>using</code>)</span>
+  </dt>
+  <dd>
+    Indicate if the payment credential (e.g. PAN and expiry) will be stored for
+    a later transaction where the payment credential is not provided by the
+    cardholder but collected from (encrypted) storage.
+    <br />
+    <code>storing</code>: The payment credential will be stored; may only be
+    stored if the authorization is approved.
+    <br />
+    <code>using</code>: The payment credential was already stored and are now
+    being used.
+    <br />
+    If <code>initiator</code> is <code>merchant</code> then
+    <code>storing</code> is invalid and <code>using</code> is assumed.
+    For a first-in-series authorization, <code>storing</code> is assumed.
+
+    <div class="type">Optional</code>
+  </dd>
+
+  <dt>initiator
+    <span class="type">(<code>cardholder</code>|<code>merchant</code>)</span>
+  </dt>
+  <dd>
+    The initiator of the authorization. An authorization is initiated by the
+    cardholder if the cardholder decided the transaction should be created.
+    This is regardless of whether stored payment credential is being used.
+    <br />
+    For compliance reasons there should be a previous approved transaction
+    marked with <code>storing</code> before <code>initiator</code> may be
+    <code>merchant</code>.
+    For subsequent-in-series authorizations, <code>merchant</code> is assumed;
+    otherwise, <code>cardholder</code> is assumed.
+
+    <div class="type">Optional</div>
+  </dd>
+
   <dt>ip
     <span class="type">[0-9.a-fA-F:]{3,45}</span>
   </dt>
@@ -553,13 +580,50 @@ Exactly one payment method must be used.
     Cardholder's IP address. It must be a valid v4 or v6 address.
     <div class="type">Optional<br></div>
   </dd>
-  <dt>recurring
-    <span class="type">(true|false)</span>
+
+  <dt>
+    series[type]
+    <span class="type">(<code>recurring</code>|<code>unscheduled</code>)</span>
   </dt>
   <dd>
-    Must be <code>true</code> for recurring payments.
-    <div class="type">Optional</div>
+    Indicate the type of series.
+    <br />
+    <code>recurring</code>: A series of transactions where the cardholder has
+    explicitly agreed that the merchant will repeatedly charge the cardholder at
+    regular, predetermined intervals.
+    <br />
+    <code>unscheduled</code>: A series of transactions where the cardholder has
+    explicitly agreed that the merchant will repeatedly charge the cardholder at
+    unknown times, e.g. based on cardholder usage.
+    <br />
+    Although the amount may vary among transactions in any series, the amount
+    of any subsequent-in-series transaction can not exceed the amount of the
+    authentication of the first-in-series authorization.
+
+    <div class="type">Conditional. Cannot be present if <code>series[previous][id]</code> is present.</div>
   </dd>
+
+  <dt>
+    series[previous][id]
+    <span class="type">[:UUIDv4:]</span>
+  </dt>
+  <dd>
+    The Clearhaus authorization ID as a reference to the latest approved
+    authorization in the series. The authorization is processed as a
+    first-in-series if not populated while <code>series[type]</code> is
+    populated.
+    <br />
+    Can be used only with payment method <code>card</code>.
+    <br />
+    If the latest approved authorization in the series was not processed via
+    Clearhaus, after obtaining explicit approval from Clearhaus, you can
+    provide raw scheme values; see
+    <a href="#scheme-reference-to-series">Scheme reference to series</a>.
+
+    <div class="type">Conditional. Cannot be present if <code>series[type]</code> is present.</div>
+  </dd>
+
+
   <dt>text_on_statement
     <span class="type">[\x20-\x7E]{2,22}
       <a target="_blank" href="http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">
@@ -572,6 +636,7 @@ Exactly one payment method must be used.
     <div class="type">May not be all digits, all same character, or all sequential characters (e.g. "abc")</div>
     <div class="type">Optional</div>
   </dd>
+
   <dt>reference
     <span class="type">[\x20-\x7E]{1,30}
       <a target="_blank" href= "http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">
@@ -582,6 +647,14 @@ Exactly one payment method must be used.
   <dd>
     A reference to an external object, such as an order number.
     <div class="type">Optional</div>
+  </dd>
+
+  <!-- deprecated -->
+  <dt><strike>recurring</strike>
+    <span class="type">(true|false)</span>
+  </dt>
+  <dd>
+    Deprecated! Please use <code>series</code>.
   </dd>
 </dl>
 
@@ -795,6 +868,58 @@ Only one 3-D Secure version can be used for a given authorization.
   <dd>
     The 3-D Secure version 2 RReq containing <code>authenticationValue</code>, <code>dsTransID</code>, etc.
     <div class="type">Optional. Cannot be present if <code>ares</code> is present.</div>
+  </dd>
+</dl>
+
+##### Scheme reference to series
+
+If the previous-in-series authorization was made via this API, you should use
+`series[previous][id]` to reference the previous-in-series authorization.
+Otherwise, you must obtain explicit approval from Clearhaus to use these raw
+scheme values grouped in <code>series[previous][mastercard]</code> and
+<code>series[previous][visa]</code>.
+
+The Mastercard specific reference to the series contains the following parts.
+
+<dl class="dl-vertical">
+  <dt>
+    series[previous][mastercard][tid]
+    <span class="type">[A-z0-9]{3}[A-z0-9]{6}[0-9]{4}[ ]{2}</span>
+  </dt>
+  <dd>
+    Trace ID being the concatenation of values
+    Data element 63 subfield 1 (Financial Network Code) (position 1-3),
+    Data element 63 subfield 2 (Banknet Reference Number) (position 4-9),
+    Data element 15 (Date, Settlement) (position 10-13), and
+    two spaces;
+    to be used in Data Element 48, Subfield 63.
+
+    <div class="type">Optional. Cannot be present if <code>series[previous][id]</code> or <code>series[previous][visa]</code> is present.</div>
+  </dd>
+
+  <dt>
+    series[previous][mastercard][tokenized]
+    <span class="type">(true|false)</span>
+  </dt>
+  <dd>
+    Indicate if the series was initiated with a tokenized payment method.
+
+    <div class="type">Conditional. Required if <code>series[previous][mastercard][tid]</code> is present.</div>
+  </dd>
+</dl>
+
+The Visa specific reference to the series has only one part.
+
+<dl class="dl-vertical">
+  <dt>
+    series[previous][visa][tid]
+    <span class="type">[0-9]{15}</span>
+  </dt>
+  <dd>
+    Transaction ID from Field 62.2 of the first (or previous) in series
+    authorization; to be used in Field 125, Usage 2, Dataset ID 03.
+
+    <div class="type">Optional. Cannot be present if <code>series[previous][id]</code> or <code>series[previous][mastercard]</code> is present.</div>
   </dd>
 </dl>
 
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -919,7 +919,7 @@ The Mastercard specific reference to the series contains the following parts.
 <dl class="dl-vertical">
   <dt>
     series[previous][mastercard][tid]
-    <span class="type">[A-z0-9]{3}[A-z0-9]{6}[0-9]{4}[ ]{2}</span>
+    <span class="type">[A-Za-z0-9]{3}[A-Za-z0-9]{6}[0-9]{4}[ ]{2}</span>
   </dt>
   <dd>
     Trace ID being the concatenation of values
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -743,7 +743,7 @@ Exactly one payment method must be used.
 <p class="alert alert-info">
   <b>Notice:</b> An authorization that includes
   <code>card[3dsecure][v1][pares]</code>, <code>card[3dsecure][v2][rreq]</code>,
-  and/or <code>card[csc]</code> cannot be a subsequent recurring authorization.
+  and/or <code>card[csc]</code> cannot be a subsequent-in-series authorization.
 </p>
 
 
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -616,6 +616,18 @@ Exactly one payment method must be used.
     <div class="type">Optional<br></div>
   </dd>
 
+  <dt>reference
+    <span class="type">[\x20-\x7E]{1,30}
+      <a target="_blank" href= "http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">
+        ASCII printable characters
+      </a>
+    </span>
+  </dt>
+  <dd>
+    A reference to an external object, such as an order number.
+    <div class="type">Optional</div>
+  </dd>
+
   <dt>
     series[type]
     <span class="type">(<code>recurring</code>|<code>unscheduled</code>)</span>
@@ -658,6 +670,7 @@ Exactly one payment method must be used.
 
     <div class="type">Conditional. Cannot be present if <code>series[type]</code> is present.</div>
   </dd>
+
   <dt>text_on_statement
     <span class="type">[\x20-\x7E]{2,22}
       <a target="_blank" href="http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">
@@ -668,18 +681,6 @@ Exactly one payment method must be used.
   <dd>
     Text that will be placed on cardholder's bank statement.
     <div class="type">May not be all digits, all same character, or all sequential characters (e.g. "abc")</div>
-    <div class="type">Optional</div>
-  </dd>
-
-  <dt>reference
-    <span class="type">[\x20-\x7E]{1,30}
-      <a target="_blank" href= "http://en.wikipedia.org/wiki/ASCII#ASCII_printable_characters">
-        ASCII printable characters
-      </a>
-    </span>
-  </dt>
-  <dd>
-    A reference to an external object, such as an order number.
     <div class="type">Optional</div>
   </dd>
 
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -1146,7 +1146,7 @@ POST https://gateway.clearhaus.com/credits
 
 <p class="alert alert-info">
 <b>Notice:</b> Implicitly, <code>initiator</code> is <code>merchant</code> and
-<code>credential</code> is <code>used</code>.
+<code>credential_on_file</code> is <code>use</code>.
 </p>
 
 ### Account
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -990,6 +990,14 @@ The Visa specific reference to the series has only one part.
   </dd>
 </dl>
 
+<p class="alert alert-info">
+  <b>Notice:</b> A series migrated to Clearhaus using these scheme references
+  cannot be continued with the now deprecated <code>recurring</code> flag.
+  Instead, the subsequent-in-series following an authorization using scheme
+  references must use <code>series[previous][id]</code> to point to the
+  previous in series.
+</p>
+
 
 ### Captures
 
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -965,19 +965,6 @@ The Mastercard specific reference to the series contains the following parts.
     Cannot be present if <code>series[previous][id]</code> or
     <code>series[previous][visa]</code> is present.</div>
   </dd>
-
-  <dt>
-    series[previous][mastercard][tokenized]
-    <span class="type">(true|false)</span>
-  </dt>
-  <dd>
-    Indicate if the series was initiated with a tokenized payment.
-
-    <div class="type">Conditional.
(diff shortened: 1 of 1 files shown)


source/index.html.md
```diff
@@ -540,26 +527,103 @@ Exactly one payment method must be used.
...
   <dd>
     <a target="_blank" href="currencies.txt">3-letter currency code</a>. (Some exponents differ from ISO 4217.)
   </dd>
+
+  <dt>credential
+    <span class="type">(<code>storing</code>|<code>using</code>)</span>
+  </dt>
+  <dd>
+    Indicate if the payment credential (e.g. PAN and expiry) will be stored for
+    a later transaction where the payment credential is not provided by the
```
mt-clearhaus, line 543: Suggestion: `s/a later transaction/future use/`

source/index.html.md
```diff
@@ -540,26 +527,103 @@ Exactly one payment method must be used.
...
+    cardholder but collected from (encrypted) storage.
+    <br />
+    <code>storing</code>: The payment credential will be stored; may only be
+    stored if the authorization is approved.
+    <br />
+    <code>using</code>: The payment credential was already stored and are now
```
mt-clearhaus, line 549:
```suggestion
    <code>using</code>: The payment credential has already been stored and is now
```

source/index.html.md
```diff
@@ -540,26 +527,103 @@ Exactly one payment method must be used.
...
+    being used.
+    <br />
+    If <code>initiator</code> is <code>merchant</code> then
+    <code>storing</code> is invalid and <code>using</code> is assumed.
```
mt-clearhaus, line 553: Suggestion to be further polished and writting using bullets: "If `initiator` is `merchant` then: a) `storing` is invalid and will cause an input error, and b) `using` is assumed if `credential` is left out."

Also, add "if `credential` is left out" to the next line (554). Of course that's also just a suggestion.

source/index.html.md
```diff
@@ -540,26 +527,103 @@ Exactly one payment method must be used.
...
+    <div class="type">Optional</code>
+  </dd>
+
+  <dt>initiator
+    <span class="type">(<code>cardholder</code>|<code>merchant</code>)</span>
+  </dt>
+  <dd>
+    The initiator of the authorization. An authorization is initiated by the
+    cardholder if the cardholder decided the transaction should be created.
+    This is regardless of whether stored payment credential is being used.
```
mt-clearhaus, line 565:
```suggestion
    This is regardless of whether a stored payment credential is being used.
```

source/index.html.md
```diff
@@ -540,26 +527,103 @@ Exactly one payment method must be used.
...
+    <br />
+    For compliance reasons there should be a previous approved transaction
+    marked with <code>storing</code> before <code>initiator</code> may be
+    <code>merchant</code>.
+    For subsequent-in-series authorizations, <code>merchant</code> is assumed;
```
mt-clearhaus, line 570:
```suggestion
    For subsequent-in-series authorizations, if <code>initiator</code> is left out, <code>merchant</code> is assumed;
```

To be in line with my previous suggestions regarding lines 553-554. I don't really like the phrase "left out" :man_shrugging:

source/index.html.md
```diff
@@ -540,26 +527,103 @@ Exactly one payment method must be used.
...
+    explicitly agreed that the merchant will repeatedly charge the cardholder at
+    regular, predetermined intervals.
+    <br />
+    <code>unscheduled</code>: A series of transactions where the cardholder has
+    explicitly agreed that the merchant will repeatedly charge the cardholder at
+    unknown times, e.g. based on cardholder usage.
+    <br />
+    Although the amount may vary among transactions in any series, the amount
+    of any subsequent-in-series transaction can not exceed the amount of the
+    authentication of the first-in-series authorization.
```
mt-clearhaus, line 601: Perhaps emphasize "authentication" somehow?

source/index.html.md
```diff
@@ -540,26 +527,103 @@ Exactly one payment method must be used.
...
+
+  <dt>
+    series[previous][id]
+    <span class="type">[:UUIDv4:]</span>
+  </dt>
+  <dd>
+    The Clearhaus authorization ID as a reference to the latest approved
+    authorization in the series. The authorization is processed as a
+    first-in-series if not populated while <code>series[type]</code> is
+    populated.
```
mt-clearhaus, line 614: I think something is wrong with this sentence?

"a is processed as b if not populated while c is populated"

I'm confused :smile:

source/index.html.md
```diff
@@ -540,26 +527,103 @@ Exactly one payment method must be used.
...
+    <br />
+    If the latest approved authorization in the series was not processed via
+    Clearhaus, after obtaining explicit approval from Clearhaus, you can
+    provide raw scheme values; see
+    <a href="#scheme-reference-to-series">Scheme reference to series</a>.
+
+    <div class="type">Conditional. Cannot be present if <code>series[type]</code> is present.</div>
   </dd>
+
+
```
mt-clearhaus, line 626:
```suggestion

```

source/index.html.md
```diff
@@ -583,6 +648,14 @@ Exactly one payment method must be used.
...
     <div class="type">Optional</div>
   </dd>
+
+  <!-- deprecated -->
+  <dt><strike>recurring</strike>
+    <span class="type">(true|false)</span>
+  </dt>
+  <dd>
+    Deprecated! Please use <code>series</code>.
+  </dd>
```
mt-clearhaus, line 658: I think we ought to update the examples here and there to use `series` rather than the deprecated `recurring`.

mt-clearhaus: E.g. in line 361.

mt-clearhaus: I understand that this doesn't make sense before the new parameters have stabilized.

source/index.html.md
```diff
@@ -798,6 +871,58 @@ Only one 3-D Secure version can be used for a given authorization.
   </dd>
 </dl>
 
+##### Scheme reference to series
+
+If the previous-in-series authorization was made via this API, you should use
+`series[previous][id]` to reference the previous-in-series authorization.
+Otherwise, you must obtain explicit approval from Clearhaus to use these raw
```
mt-clearhaus, line 878:
```suggestion
Otherwise, you must obtain explicit approval from Clearhaus to use the raw
```

source/index.html.md
```diff
@@ -798,6 +871,58 @@ Only one 3-D Secure version can be used for a given authorization.
...
+scheme values grouped in <code>series[previous][mastercard]</code> and
+<code>series[previous][visa]</code>.
+
+The Mastercard specific reference to the series contains the following parts.
+
+<dl class="dl-vertical">
+  <dt>
+    series[previous][mastercard][tid]
+    <span class="type">[A-z0-9]{3}[A-z0-9]{6}[0-9]{4}[ ]{2}</span>
```
mt-clearhaus, line 887: In other places we use `A-Za-z`.

source/index.html.md
```diff
@@ -798,6 +871,58 @@ Only one 3-D Secure version can be used for a given authorization.
...
+
+    <div class="type">Optional. Cannot be present if <code>series[previous][id]</code> or <code>series[previous][visa]</code> is present.</div>
+  </dd>
+
+  <dt>
+    series[previous][mastercard][tokenized]
+    <span class="type">(true|false)</span>
+  </dt>
+  <dd>
+    Indicate if the series was initiated with a tokenized payment method.
```
mt-clearhaus, line 905: We might need to add some additional explanation.

mt-clearhaus (commented):
I have left some comments, however, nice work :+1:

source/index.html.md
ct-clearhaus:
```suggestion
    future use where the payment credential is not provided by the
```

source/index.html.md
ct-clearhaus: I am in doubt if we want to word it like "using a stored credential" or "using stored credential". Looking at Visa, they use both; perhaps there's some systematism that I just don't get?

source/index.html.md
ct-clearhaus: Good thinking.

I initially emphasized "can not exceed", but discarded that because I thought that just about every other word in the sentence is equally important. Dunno, really ... 🙂

If you give it a second thought and agree, then please feel free to push or drop a "PR suggestion" — I'll do the same if you don't beat me to it 🙂

source/index.html.md
ct-clearhaus: I see that this became confusingly minimalistic.

Let me try to "de-confuse" 😄

> "a is processed as b if <this field, implicitly> [is] not populated while [in the sense "but"] c is populated"

I'll try to fix the wording 🙂

source/index.html.md
ct-clearhaus: I'll fix this 🙂

source/index.html.md
ct-clearhaus: Nice catch!

I preferred the shorter one here because the regex is already pretty long 🙁

WDYT? Should we adjust the rest or this one? Or keep it inconsistent (me not like)?

source/index.html.md
ct-clearhaus: I didn't think about that, but you're right, we definitely do!

I'm in doubt if it should it go here or in a guideline — WDYT?

ct-clearhaus (commented):
Thanks a lot for the early 👀!

source/index.html.md
```diff
@@ -540,25 +554,102 @@ Exactly one payment method must be used.
...
+
   <dt>currency
     <span class="type">[A-Z]{3}</span>
   </dt>
   <dd>
     <a target="_blank" href="currencies.txt">3-letter currency code</a>. (Some exponents differ from ISO 4217.)
   </dd>
+
+  <dt>credential
+    <span class="type">(<code>storing</code>|<code>using</code>)</span>
```
ct-clearhaus, line 566:
```suggestion
    <span class="type">(<code>storing</code>|<code>from_storage</code>)</span>
```
WDYT @clearhaus/pci?

source/index.html.md
ct-clearhaus: Fixed by e4a6aebaf5e53b32346e1cb0199460b395865f85

source/index.html.md
ct-clearhaus: Fixed by https://github.com/clearhaus/gateway-api-docs/pull/111/commits/e4a6aebaf5e53b32346e1cb0199460b395865f85#diff-d30d91dbcd59693b69f9242acb4630c5R601-R606

source/index.html.md
ct-clearhaus: Fixed by https://github.com/clearhaus/gateway-api-docs/pull/111/commits/e4a6aebaf5e53b32346e1cb0199460b395865f85#diff-d30d91dbcd59693b69f9242acb4630c5R647-R650

source/index.html.md
ct-clearhaus: Was fixed by https://github.com/clearhaus/gateway-api-docs/pull/111/commits/87a8acf89466d7c1d3d8a1ea4f0cd244630e379c#diff-d30d91dbcd59693b69f9242acb4630c5R364

source/index.html.md
mt-clearhaus: I think my suggestion was prompted by "**the** payment credential" in many of the above lines.

I don't think there is any systematism about this in scheme documentation, however, I don't know for sure.

source/index.html.md
```diff
@@ -540,26 +554,123 @@ Exactly one payment method must be used.
...
+    explicitly agreed that the merchant will repeatedly charge the cardholder at
+    regular, predetermined intervals that may not exceed 1 year.
+    <br />
+    <code>unscheduled</code>: A series of transactions where the cardholder has
+    explicitly agreed that the merchant will repeatedly charge the cardholder at
+    unknown times, e.g. based on cardholder usage.
+    <br />
+    Although the amount may vary among transactions in any series, the amount
+    of any subsequent-in-series transaction can not exceed the amount of the
+    authentication of the first-in-series authorization.
```
mt-clearhaus, line 648:
```suggestion
    of any subsequent-in-series transaction may not exceed the amount of the
    <b>authentication</b> of the first-in-series authorization.
```

source/index.html.md
mt-clearhaus: https://stackoverflow.com/a/4923397

```
> /[A-z]/ =~ ']'
=> 0
```

source/index.html.md
```diff
@@ -798,6 +906,58 @@ Only one 3-D Secure version can be used for a given authorization.
...
+Otherwise, you must obtain explicit approval from Clearhaus to use the raw
+scheme values grouped in <code>series[previous][mastercard]</code> and
+<code>series[previous][visa]</code>.
+
+The Mastercard specific reference to the series contains the following parts.
+
+<dl class="dl-vertical">
+  <dt>
+    series[previous][mastercard][tid]
+    <span class="type">[A-z0-9]{3}[A-z0-9]{6}[0-9]{4}[ ]{2}</span>
```
mt-clearhaus, line 922:
```suggestion
    <span class="type">[A-Za-z0-9]{3}[A-Za-z0-9]{6}[0-9]{4}[ ]{2}</span>
```

source/index.html.md
mt-clearhaus: I lean towards putting it in a guideline rather than having it here.

Are we sure that "tokenized payment method" is correct wording? For instance, Google Pay, a payment method, can be both tokenized and non-tokenized :man_shrugging:

source/index.html.md
mt-clearhaus: `from_storage` might be a tad better than `using` because we have chosen `credential` rather than `cof`.

source/index.html.md
```diff
@@ -540,26 +554,123 @@ Exactly one payment method must be used.
...
   </dd>
+
   <dt>currency
     <span class="type">[A-Z]{3}</span>
   </dt>
   <dd>
     <a target="_blank" href="currencies.txt">3-letter currency code</a>. (Some exponents differ from ISO 4217.)
   </dd>
+
+  <dt>credential
```
ct-clearhaus, line 565:
```suggestion
  <dt>cof
```

source/index.html.md
ct-clearhaus: Let's "unfold" the `cof` stuff over there: https://github.com/clearhaus/gateway-api-docs/pull/111/files#r497447364

source/index.html.md
ct-clearhaus: 😄 I actually wrote `may` first, but ... oh well ... https://english.stackexchange.com/questions/56065/must-not-or-may-not-which-is-the-most-correct

We should probably use "RFC style" wording — and reference https://tools.ietf.org/html/rfc2119

source/index.html.md
ct-clearhaus: > Are we sure "tokenized payment method" is correct wording? For instance, Google Pay, a payment method, can be both tokenized and non-tokenized 🤷‍♂️

Luckily the current documented version (on `master`) doesn't support Google Pay 😄

Though, I think we should just word it like this:
```suggestion
    Indicate if the series was initiated with a tokenized payment.
```

source/index.html.md
mt-clearhaus: I also like "must", perhaps better than "may", however, I think "cannot" isn't quite correct :slightly_smiling_face:

source/index.html.md
mt-clearhaus: I think that works well :+1:

source/index.html.md
```diff
@@ -540,26 +554,123 @@ Exactly one payment method must be used.
...
+    explicitly agreed that the merchant will repeatedly charge the cardholder at
+    regular, predetermined intervals that may not exceed 1 year.
+    <br />
+    <code>unscheduled</code>: A series of transactions where the cardholder has
+    explicitly agreed that the merchant will repeatedly charge the cardholder at
+    unknown times, e.g. based on cardholder usage.
+    <br />
+    Although the amount may vary among transactions in any series, the amount
+    of any subsequent-in-series transaction can not exceed the amount of the
+    authentication of the first-in-series authorization.
```
mt-clearhaus, line 648:
```suggestion
    of any subsequent-in-series transaction must not exceed the amount of the
    <b>authentication</b> of the first-in-series authorization.
```

source/index.html.md
ad-clearhaus: Are we going to enforce this? Can not (can) imply that we return an error if you try. Must/may not more has a "it is on you to secure".

mt-clearhaus: @ad-clearhaus Although I don't agree with your interpretation of the words, I think whether or not we will enforce the rule is a very relevant question.

ct-clearhaus: We cannot (`:tøhø:`) enforce it because we are blind about the authentication amount.

It is a compliance-wise "it wrong if you do" — if scheme or issuer declines, I have no clue, but they ought to decline.

Perhaps we should move this entire section into the guideline?

source/index.html.md
ct-clearhaus: Someone can interpret this as "it doesn't *have* to exceed ..." 😜

source/index.html.md
ad-clearhaus: Maybe this should indeed be moved to guidelines.
We are not going to enforce it. The issuers should enforce it and decline. Which is fine. Then we can return that to the PSP's. As it stands we have no transaction evidence pointing towards the issuers actually declining.
To me, this issue seems like something where we need to see over time how the ecosystem reacts.

But I still think, in guidelines or not, that _cannot_ should be replaced by _must_ ;)

source/index.html.md
ct-clearhaus: In no specific order:
1. `cof = (first|subsequent)`
2. `cof = (store|use)`
3. `cof = (storing|using)`

mt-clearhaus: I like the simplicity/brevity of 2, and I find the phrases "store credential of file" and "use credential on file" meaningful and clear :slightly_smiling_face:

ct-clearhaus: I have a 🥩 with `cof` — that it's a TLA that we need to explain. That's why I tried to come up with something that could compete, and my best attempt was if we are `storing [the] credential` or `[grabbing the] credential from_storage` (with `using` I was trying to communicate `using [stored] credential`, but the `stored` piece is pretty important word to describe the right intent).

source/index.html.md
mt-clearhaus: I can accept "cannot" :stuck_out_tongue: 

Alternatively, "must be less than or equal to the amount of the" :smile:

source/index.html.md
ct-clearhaus: Resolved by 8415d33.

source/index.html.md
```diff
@@ -540,26 +554,119 @@ Exactly one payment method must be used.
...
+
   <dt>currency
     <span class="type">[A-Z]{3}</span>
   </dt>
   <dd>
     <a target="_blank" href="currencies.txt">3-letter currency code</a>. (Some exponents differ from ISO 4217.)
   </dd>
+
+  <dt>credential
+    <span class="type">(<code>storing</code>|<code>using</code>)</span>
```
ct-clearhaus, line 566: Two points:

1. We use `text_on_statement` to avoid the TLA `tos`; likewise, we should just use `credential_on_file`.
2. It is hard not to like the simplicity/brevity of `store|use` (courtesy @mt-clearhaus, https://github.com/clearhaus/gateway-api-docs/pull/111/files#r500507071).

Thus:

```suggestion
  <dt>credential_on_file
    <span class="type">(<code>store</code>|<code>use</code>)</span>
```

In return I have marked https://github.com/clearhaus/gateway-api-docs/pull/111/files#r497447364 and https://github.com/clearhaus/gateway-api-docs/pull/111/files#r497425813 as resolved.

source/index.html.md
ct-clearhaus: Migrated to https://github.com/clearhaus/gateway-api-docs/pull/111/files#r524296900.

source/index.html.md
ct-clearhaus: Migrated to https://github.com/clearhaus/gateway-api-docs/pull/111/files#r524296900.

source/index.html.md
```diff
@@ -540,26 +554,119 @@ Exactly one payment method must be used.
...
+    <br />
+    <code>using</code>: The payment credential has already been stored and is now
+    being used.
+    <br />
+    Default:
+    <ul>
+      <li><code>using</code>, if <code>initiator</code> is <code>merchant</code>
+      (<code>storing</code> is invalid),</li>
+      <li><code>storing</code>, if the authorization is first-in-series.</li>
+    </ul>
```
ct-clearhaus, line 584: Consequence of https://github.com/clearhaus/gateway-api-docs/pull/111/files#r524296900.

```suggestion
    <code>store</code>: The payment credential will be stored; may only be
    stored if the authorization is approved.
    <br />
    <code>use</code>: The payment credential has already been stored and is now
    being used.
    <br />
    Default:
    <ul>
      <li><code>use</code>, if <code>initiator</code> is <code>merchant</code>
      (<code>store</code> is invalid),</li>
      <li><code>store</code>, if the authorization is first-in-series.</li>
    </ul>
```

source/index.html.md
```diff
@@ -333,64 +333,78 @@ Depending on card scheme and merchant category, the name on the card might be
...
 
-This should be followed by a capture except when the amount is `0`.
+This should be followed by a capture.
 
-Subsequent authorizations are made similarly, however, CSC would not be
-included. For 3-D Secure version 1 (see [3-D Secure](#3-d-secure)), PARes would
-not be included. 3-D Secure version 2 has some support for subsequent
-authorizations.
+Subsequent recurring authorizations initiated by the merchant are made
+similarly, however, CSC would not be included, and the previous-in-series is
```
mt-clearhaus, line 385:
```suggestion
similarly, however, CSC is not included, and the previous-in-series is
```

source/index.html.md
```diff
@@ -540,26 +554,119 @@ Exactly one payment method must be used.
...
+
+  <dt>credential_on_file
+    <span class="type">(<code>store</code>|<code>use</code>)</span>
+  </dt>
+  <dd>
+    Indicate if the payment credential (e.g. PAN and expiry) will be stored for
+    future use where the payment credential is not provided by the
+    cardholder but collected from (encrypted) storage.
+    <br />
+    <code>store</code>: The payment credential will be stored; may only be
```
mt-clearhaus, line 573:
```suggestion
    <code>store</code>: The payment credential will be stored; it may only be
```

source/index.html.md
```diff
@@ -981,6 +1144,11 @@ POST https://gateway.clearhaus.com/credits
   </dd>
 </dl>
 
+<p class="alert alert-info">
+<b>Notice:</b> Implicitly, <code>initiator</code> is <code>merchant</code> and
+<code>credential</code> is <code>used</code>.
```
ct-clearhaus, line 1149: Consequence:

```suggestion
<code>credential_on_file</code> is <code>use</code>.
```

source/index.html.md
```diff
@@ -933,21 +933,29 @@ The Mastercard specific reference to the series contains the following parts.
...
     series[previous][mastercard][exemption]
-    <span class="type">(<code>mit</code>|<code>recurring</code>)</span>
+    <span class="type">(<code>fixed_amount_series</code>|<code>variable_amount_series</code>)</span>
   </dt>
   <dd>
-    The <i>Mastercard exemption</i> to be used for the series. If the
-    previous-in-series is a subsequent-in-series it should be equal to the
-    <i>Mastercard exemption</i> applied for the previous-in-series.
+    The <i>Mastercard exemption</i> to be used for the series indicating if the
+    series is a fixed-amount or a variable-amount series.
```
mt-clearhaus, line 940:
```suggestion
    The <i>Mastercard exemption</i> is used to indicate if the series is 
    fixed-amount or a variable-amount.
```

source/index.html.md
```diff
@@ -333,64 +333,71 @@ Depending on card scheme and merchant category, the name on the card might be
 necessary for approval of credits. It may be provided through the optional
 parameter `card[name]`.
 
-## Recurring payments
+## Series of transactions
 
-Recurring payments enable you to repeatedly charge cardholders without providing
-CSC or other means of cardholder authentication for subsequent payments.
+Clearhaus currently support only one type of subscription billing:
```
mt-clearhaus, line 338:
```suggestion
Clearhaus currently supports only one type of subscription billing:
```

source/index.html.md
```diff
@@ -333,64 +333,71 @@ Depending on card scheme and merchant category, the name on the card might be
...
 
-### Subscription concept
+* Recurring: Transactions processed at predetermined, regular intervals not
+    exceeding 1 year; e.g. a monthly subscription for a magazine.
```
mt-clearhaus, line 341: Perhaps make it explicitly clear if a 12 month interval is ok or not.

source/index.html.md
```diff
@@ -333,64 +333,71 @@ Depending on card scheme and merchant category, the name on the card might be
...
-authorizations. Clearhaus supports subscriptions in the form of recurring
-payments.
+There may be an agreed end of the series.
 
 ### Repeatedly reserve money
 
-A recurring payment is made by making an authorization and setting `recurring`
-parameter to `true`. The first recurring payment for a given card could be made
-this way (notice that the amount may be zero):
+A first-in-series recurring payment is made by making an authorization and
```
mt-clearhaus, line 347:
```suggestion
A first-in-series recurring payment is initiated by making an authorization and
```

source/index.html.md
```diff
@@ -333,64 +333,71 @@ Depending on card scheme and merchant category, the name on the card might be
...
 
-This should be followed by a capture except when the amount is `0`.
+This should be followed by a capture.
 
-Subsequent authorizations are made similarly, however, CSC would not be
-included. For 3-D Secure version 1 (see [3-D Secure](#3-d-secure)), PARes would
-not be included. 3-D Secure version 2 has some support for subsequent
-authorizations.
+Subsequent recurring authorizations initiated by the merchant are made
+similarly, however, CSC is not included, and the previous-in-series is
```
mt-clearhaus, line 378:
```suggestion
Subsequent-in-series recurring authorizations initiated by the merchant are
made similarly, however, CSC is not included, and the previous-in-series is
```

source/index.html.md
```diff
@@ -333,64 +333,71 @@ Depending on card scheme and merchant category, the name on the card might be
...
+  -d "amount=2050"    \
+  -d "currency=EUR"   \
+  -d "series[previous][id]=1b722683-92ad-4c6b-85da-e119d550670d" \
+  -d "card[pan]=4111111111111111" \
+  -d "card[expire_month]=06"      \
+  -d "card[expire_year]=2026"     \
+  -H "Signature: <signing-api-key> RS256-hex <signature>"
+````
+
+A first-in-series authorization can also be made using the `applepay`
```
mt-clearhaus, line 394: I suppose Samsung Pay, Google Pay and MPO as well?

source/index.html.md
```diff
@@ -333,64 +333,71 @@ Depending on card scheme and merchant category, the name on the card might be
...
+payment method; subsequent-in-series authorizations, however,
+must be made using the `card` payment method using the card details of the
+previous-in-series authorization referenced.
```
mt-clearhaus, line 397:
```suggestion
referenced previous-in-series authorization.
```

source/index.html.md
```diff
@@ -540,26 +547,115 @@ Exactly one payment method must be used.
...
+    This is regardless of whether a stored payment credential is being used.
+    <br />
+    For compliance reasons there should be a previous approved transaction (for
+    the cardholder and the merchant) marked with <code>storing</code> before
+    <code>initiator</code> may be <code>merchant</code>.
+    <br />
+    Default:
+    <ul>
+      <li><code>merchant</code>, if the authorization is
+      subsequent-in-subsequent (<code>cardholder</code> is invalid),</li>
```
mt-clearhaus, line 597:
```suggestion
      subsequent-in-series recurring (<code>cardholder</code> is invalid),</li>
```

source/index.html.md
```diff
@@ -540,26 +547,115 @@ Exactly one payment method must be used.
...
+    A reference to an external object, such as an order number.
     <div class="type">Optional</div>
   </dd>
+
+  <dt>
+    series[type]
+    <span class="type"><code>recurring</code></span>
+  </dt>
+  <dd>
+    Indicate the type of series.
```
mt-clearhaus, line 629:
```suggestion
    The type of series.
```

source/index.html.md
```diff
@@ -540,26 +547,115 @@ Exactly one payment method must be used.
...
+    <br />
+    <code>recurring</code>: A series of transactions where the cardholder has
+    explicitly agreed that the merchant will repeatedly charge the cardholder at
```
mt-clearhaus, line 632:
```suggestion
    explicitly agreed that the merchant may repeatedly charge the cardholder at
```

source/index.html.md
```diff
@@ -540,26 +547,115 @@ Exactly one payment method must be used.
...
+    regular, predetermined intervals that may not exceed 1 year.
```
mt-clearhaus, line 633: I've also read this above. Perhaps remove the redundancy.

mt-clearhaus: The part about not exceeding 1 year.

source/index.html.md
```diff
@@ -540,26 +547,115 @@ Exactly one payment method must be used.
...
+
+    <div class="type">Conditional. Cannot be present if <code>series[previous]</code> is present.</div>
```
mt-clearhaus, line 635:
```suggestion
    <div class="type">
      Conditional. Cannot be present if <code>series[previous]</code> is
      present. This means that the type of an existing series cannot be
      changed.
    </div>
```

source/index.html.md
```diff
@@ -572,19 +668,27 @@ Exactly one payment method must be used.
...
+    <div class="type">Optional. Cannot be used with <code>series</code> or
+    <code>initiator</code>.</div>
   </dd>
 </dl>
 
+<p class="alert alert-info">
+  <b>Notice:</b> When <code>recurring</code> is used, Clearhaus automatically
+  identify if there was a previous-in-series and together with the level of
+  authentication (CSC, 3-D Secure, etc.) conclude if the payment is a first
+  recurring or a subsequent recurring.
```
mt-clearhaus, line 688:
```suggestion
  <b>Notice:</b> When <code>recurring</code> is used, Clearhaus automatically
  identifies if there was a previous-in-series and if that is the case uses the
  level of authentication (CSC, 3-D Secure, etc.) to conclude if the payment is
  a first-in-in-series or a subsequent-in-series recurring.
```

source/index.html.md
```diff
@@ -798,6 +902,88 @@ Only one 3-D Secure version can be used for a given authorization.
   </dd>
 </dl>
 
+##### Scheme reference to series
+
+If the previous-in-series authorization was made via this API, you should use
+`series[previous][id]` to reference the previous-in-series authorization.
+Otherwise, you must obtain explicit approval from Clearhaus to use the raw
+scheme values grouped in <code>series[previous][mastercard]</code> and
+<code>series[previous][visa]</code>.
```
mt-clearhaus, line 911:
```suggestion
If the previous-in-series authorization was made via this API, you must use
`series[previous][id]` to reference it. If it was not made via this API, you 
must obtain explicit approval from Clearhaus to use the raw scheme values
grouped in <code>series[previous][mastercard]</code> and
<code>series[previous][visa]</code>. This is relevant when moving subscriptions
to Clearhaus from another acquirer.
```

source/index.html.md
```diff
@@ -798,6 +902,88 @@ Only one 3-D Secure version can be used for a given authorization.
...
+
+The Mastercard specific reference to the series contains the following parts.
+
+<dl class="dl-vertical">
+  <dt>
+    series[previous][mastercard][tid]
+    <span class="type">[A-Za-z0-9]{3}[A-Za-z0-9]{6}[0-9]{4}[ ]{2}</span>
```
mt-clearhaus, line 918:
```suggestion
    <span class="type">[A-Za-z0-9]{3}[A-Za-z0-9]{6}[0-9]{4} {2}</span>
```

I think the financial network code and Banknet reference number are always upper case, but I'm not completely certain.

source/index.html.md
```diff
@@ -798,6 +902,88 @@ Only one 3-D Secure version can be used for a given authorization.
...
+    The <i>Mastercard exemption</i> to be used for the series indicating if the
+    series is a fixed-amount or a variable-amount series.
+    <ul>
+      <li><code>fixed_amount_series</code>:
+        The series is a fixed-amount series. (MPMI value <code>03</code>.)
+      </li>
+      <li><code>variable_amount_series</code>:
+        The series is a variable-amount series. (MPMI value <code>01</code>.)
+        A <i>Mastercard exemption</i> (not formally an acquirer exemption for
+        SCA) indicating that the transaction is out of scope for SCA.
```
mt-clearhaus, line 948: Uhm, but MIT **is** "formally" a Mastercard acquirer exemption for SCA?

mt-clearhaus: It is not for Visa.

source/index.html.md
```diff
@@ -798,6 +902,88 @@ Only one 3-D Secure version can be used for a given authorization.
...
+      </li>
+    </ul>
+    If the previous-in-series is a subsequent-in-series it should be equal to
+    the <i>Mastercard exemption</i> applied for the previous-in-series.
+    The value originates from Mastercard Data element 48, Subelement 22,
+    Subfield 1 named "Multi-Purpose Merchant Indicator (MPMI).
+    The value <code>01</code> indicates variable-amount whereas <code>01</code>
+    indicates fixed-amount.
```
mt-clearhaus, line 956:
```suggestion
    The value <code>01</code> indicates variable amount whereas <code>01</code>
    indicates fixed amount.
```

source/index.html.md
```diff
@@ -798,6 +902,88 @@ Only one 3-D Secure version can be used for a given authorization.
...
+
+    <br />
+    Clearhaus will use the same <i>Mastercard exemption</i> for the entire
+    series when later referring to the previous-in-series via
+    <code>series[previous][id]</code>.
```
mt-clearhaus, line 961:
```suggestion
    Clearhaus uses the same <i>Mastercard exemption</i> for an entire series
    when a previous-in-series authorization in the series is referenced via
    <code>series[previous][id]</code>.
```

source/index.html.md
```diff
@@ -798,6 +902,88 @@ Only one 3-D Secure version can be used for a given authorization.
...
+
+The Visa specific reference to the series has only one part.
+
+<dl class="dl-vertical">
+  <dt>
+    series[previous][visa][tid]
+    <span class="type">[0-9]{15}</span>
+  </dt>
+  <dd>
+    Transaction ID from Field 62.2 of the first (or previous) in series
```
mt-clearhaus, line 978:
```suggestion
    Transaction ID from Field 62.2 of the first-in-series or previous-in-series
```

source/index.html.md
```diff
@@ -981,6 +1167,11 @@ POST https://gateway.clearhaus.com/credits
   </dd>
 </dl>
 
+<p class="alert alert-info">
+<b>Notice:</b> Implicitly, <code>initiator</code> is <code>merchant</code> and
+<code>credential_on_file</code> is <code>use</code>.
+</p>
+
```
mt-clearhaus, line 1174: Aah, for a capture?

mt-clearhaus (commented):
I've left some suggestions, comments and questions :slightly_smiling_face:

source/index.html.md
ad-clearhaus: At some point we will expand bullet-list in the `Series of transactions` section to also include UCOF and all that jazz. So lets remove the recurring?

mt-clearhaus: Good point :+1:

source/index.html.md
mt-clearhaus: https://erinwrightwriting.com/singular-or-plural-pronouns-for-a-company/

> British English generally treats collective nouns as plural rather than singular. So, if you follow British English instead of American English, you can usually use the plural third-person pronouns *they*, *them*, *their*, and *themselves* for an individual company.

However, I believe we use American English in our code and documentation (e.g. authorization, not authorisation).

mt-clearhaus: And let me add "argh, wtf, why are there so significant differences between British and American English!?" :cry:

source/index.html.md
```diff
@@ -798,6 +902,88 @@ Only one 3-D Secure version can be used for a given authorization.
...
 
+##### Scheme reference to series
+
+If the previous-in-series authorization was made via this API, you should use
+`series[previous][id]` to reference the previous-in-series authorization.
+Otherwise, you must obtain explicit approval from Clearhaus to use the raw
+scheme values grouped in <code>series[previous][mastercard]</code> and
+<code>series[previous][visa]</code>.
+
+The Mastercard specific reference to the series contains the following parts.
```
ad-clearhaus, line 913: Flip order to start with Visa?

source/index.html.md
ct-clearhaus: 🤔 I am not even sure we know for sure; WDYT @AlexanderIversen?

source/index.html.md
ct-clearhaus: It's just that "initiated" is now getting some kind of meaning — "initiating" a transaction is now smelling like "the party that initiated it — CIT or MIT" ... WDYT?

source/index.html.md
ct-clearhaus: Regarding Google Pay and Samsung Pay: Only after https://github.com/clearhaus/gateway-api-docs/pull/69 and https://github.com/clearhaus/gateway-api-docs/pull/103 (resp.) has been merged 🙂

MPO does not allow recurring on card payments — they have MPO subscriptions and that's a different kind of beast:
* https://developer.mobilepay.dk/subscriptions-main
* https://www.mobilepay.dk/erhverv/abonnementer%20og%20fakturering/mobilepay%20subscriptions

source/index.html.md
ct-clearhaus: But still "subsequent-in-subsequent" is nonsense, so lemme suggest ...

```suggestion
      subsequent-in-series (<code>cardholder</code> is invalid),</li>
```

source/index.html.md
ct-clearhaus: I agree that we should explicitly mention that "This means that the type of an existing series cannot be changed.", however, I am not sure it belongs here. Perhaps some "Notice block" somewhere below? WDYT?

source/index.html.md
ct-clearhaus: I've tried to alphabetize 🤷‍♂️ WDYT?

source/index.html.md
ct-clearhaus: > I think the financial network code and Banknet reference number are always upper case, but I'm not completely certain.

Agreed. However, can we actually know for sure? @AlexanderIversen, do you know?

source/index.html.md
ct-clearhaus: Well, Mastercard does not mention MPMI = `01` as an exemption, but as an indication that it's out of scope — but they're also using the word "exemption" when talking about the MPMI field itself 🤷‍♂️

I'm understanding it as if they're being veeery slippery and working around PSD2 RTS regarding the requirement for SCA for amount adjustments and fixed amount recurring (and the internal conflict in the RTS that payment instruments may not be discriminated) 😃
Visa took it by the horns and kinda just allows themselves to disagree with the RTS.

I am definitely nowhere near quadratic on how we should word this. I am very open for suggestions 🙂

source/index.html.md
ct-clearhaus: I firstly wrote it as is (with dashes), then went back and removed them (because it's wrong), but then I figured that it maybe confuse some or maybe make it harder to search, so I inserted the dashes again. But since you noticed, let's remove them 😄

source/index.html.md
ct-clearhaus: It is so incredibly nice that you stumble upon this and correct it! It's just so much better. Thanks! 🙇‍♂️

source/index.html.md
ct-clearhaus: Fro a credit ... or ... am I misunderstanding something?

source/index.html.md
AlexanderIversen: From Mastercard Transaction Processing Rules I have the following quote: 

> Use may occur periodically, such as on a monthly, quarterly, or
annual basis, or as needed to “top up” the Cardholder’s account with the
Merchant (Transaction Processing Rules, 13 August 2020, p. 158)

This is not very specific but I agree that intervals should not exceed 1 year (1 year and 1 day) but I agree with @mt-clearhaus that an interval can be 1 year is OK. 

From Visa Core Rules I have the following: 

> A Transaction in a series of Transactions that use a Stored Credential
and that are processed at fixed, regular intervals (not to exceed one
year between Transactions), representing Cardholder agreement for
the Merchant to initiate future Transactions for the purchase of goods
or services provided at regular intervals. (Visa Core Rules, 18 August 2020, p. 1126)

Visas statement is actually more precise. They state that intervals cannot exceed 1 year between transactions.

ct-clearhaus: Leap years? 😆

ct-clearhaus:
```suggestion
    exceeding 366 days; e.g. a monthly subscription for a magazine.
```
WDYT?

source/index.html.md
mt-clearhaus: I see your point. I only wanted something else than "is made by making" :slightly_smiling_face: 

Perhaps just "is created by making"?

source/index.html.md
mt-clearhaus: I was surprised to read about MPMI values here. Are they really necessary? Does the audience need to know? I haven't (yet) read about them anywhere else, and I think I've read a fair amount of PSD2 SCA documents. If we zoom out, a number of SCA acquirer exemptions exist, and some of them are shared by both schemes, and some of them are only for one of them. AFAICS, MIT is an exemption only for Mastercard. Visa just says MIT is out of scope rather than define an exemption. In this context I don't understand the parenthesis in "A Mastercard exemption (not formally an acquirer exemption for SCA) [...]", and my reaction was actually only to related to that.

But it's certainly a minefield :slightly_smiling_face:

source/index.html.md
mt-clearhaus: I like it :smile: :+1: Because of `>` rather than `>=` it allows for leap years.

mt-clearhaus: But then, uhm, permaybehaps "exceeding 1 year" isn't too bad after all :thinking: 

When reading it the first time, I wasn't sure it would be ok to charge once a year, but the second time around I find it "not that bad".

source/index.html.md
mt-clearhaus: A notice block would be fine :+1: I just think it ought to be somewhere. Perhaps it fits somewhat better in a notice block rather than here.

source/index.html.md
mt-clearhaus: Right! This is placed under Credits. I'm sorry about the noise.

source/index.html.md
ad-clearhaus: > I've tried to alphabetize 

That was my hunch. :smile: 

Maybe I am colored by my surroundings. But we usually focus on Visa first for several reasons:
* Visa is the biggest of the two schemes globally
* Visa is bigger in Clearhaus' portfolio
* Mastercard is very small in DK, and we still have a DK market bias
* Historically Visa has taken the lead and Mastercard has followed after changing the namingscheme

Am I wrong?

source/index.html.md
ct-clearhaus:
```suggestion
A first-in-series recurring payment is created by making an authorization and
```

source/index.html.md
ct-clearhaus: Shortly regarding MPMI: We haven't thought about this issue before, it was only now I realized that the consequences of a subsequent-in-series being marked with MPMI `01` or `03`.

If a previous-in-series created via another acquirer was a first-in-series, then the PSP should choose if the current subsequent-in-series (the first subsequent in series, because the previous was a first-in-series) should allow for varying amount or not.

If a previous-in-series created via another acquirer was a subsequent-in-series, then the PSP should make the MPMI match the one used for the previous-in-series. Otherwise, the series change between being a varying-amount and fixed-amount series.

source/index.html.md
ct-clearhaus: You are right AFAICT. (On the last bullet, I guess it depends on the context.) Feel free to push 🙂

source/index.html.md
ct-clearhaus: I think I'll keep the original text and just allow it to be imprecise. If necessary we'll elaborate in the guideline or fix if enough people are asking 🙂

source/index.html.md
mt-clearhaus: Argh. Indeed not simple, but it makes good sense then. I assume this is or will be a part of our separate guide.

source/index.html.md
```diff
@@ -798,6 +902,89 @@ Only one 3-D Secure version can be used for a given authorization.
...
+<dl class="dl-vertical">
+  <dt>
+    series[previous][mastercard][tid]
+    <span class="type">[A-Za-z0-9]{3}[A-Za-z0-9]{6}[0-9]{4} {2}</span>
+  </dt>
+  <dd>
+    Trace ID being the concatenation of values
+    Data element 63 subfield 1 (Financial Network Code) (position 1-3),
+    Data element 63 subfield 2 (Banknet Reference Number) (position 4-9),
+    Data element 15 (Date, Settlement) (position 10-13), and
```
mt-clearhaus, line 925:
```suggestion
    Data element 15 (Date, Settlement, in MMDD format) (position 10-13), and
```

Source: Mastercard's Customer Interface Specification

source/index.html.md
ct-clearhaus: 436a301

source/index.html.md
```diff
@@ -572,19 +668,32 @@ Exactly one payment method must be used.
...
+<p class="alert alert-info">
+  <b>Notice:</b> When <code>recurring</code> is used, Clearhaus automatically
+  identifies if there was a previous-in-series and if that is the case uses the
+  level of authentication (CSC, 3-D Secure, etc.) to conclude if the payment is
+  a first-in-in-series or a subsequent-in-series recurring.
+</p>
+
+<p class="alert alert-info">
+  <b>Notice:</b> Since <code>series[type]</code> cannot be supplied together
+  with <code>series[previous]</code>, a series will stay of the same type.
```
mt-clearhaus, line 693:
```suggestion
  with <code>series[previous]</code>, the type of a series cannot change.
```

mt-clearhaus: Of course: :golfing: :man_shrugging:

source/index.html.md
```diff
@@ -798,6 +907,97 @@ Only one 3-D Secure version can be used for a given authorization.
...
+    Cannot be present if <code>series[previous][id]</code> or
+    <code>series[previous][mastercard]</code> is present.</div>
+  </dd>
+</dl>
+
+<p class="alert alert-info">
+  <b>Notice:</b> A series migrated to Clearhaus using these scheme references
+  cannot be continued with the now deprecated <code>recurring</code> flag.
+  Instead, the subsequent-in-series following an authorization using scheme
+  references must use <code>series[previous][id]</code> to point to the
```
mt-clearhaus, line 997:
```suggestion
  Instead, the subsequent-in-series following an authorization created using
  scheme references must use <code>series[previous][id]</code> to point to the
```

source/index.html.md
mt-clearhaus: (in line 341)

ct-clearhaus: Fixed in ada4b57.

source/index.html.md
```diff
@@ -333,64 +333,69 @@ Depending on card scheme and merchant category, the name on the card might be
 necessary for approval of credits. It may be provided through the optional
 parameter `card[name]`.
 
-## Recurring payments
+## Series of transactions
 
-Recurring payments enable you to repeatedly charge cardholders without providing
-CSC or other means of cardholder authentication for subsequent payments.
+Clearhaus currently supports the recurring type of subscription billing.
```
ad-clearhaus, line 338:
```suggestion
Clearhaus supports the recurring type of subscription billing.
```
After restructuring away from the enumeration list, _only_ has become an unnecessary filler

ad-clearhaus (approved):
Looks good. :+1:

source/index.html.md
```diff
@@ -333,64 +333,69 @@ Depending on card scheme and merchant category, the name on the card might be
...
+  -H "Signature: <signing-api-key> RS256-hex <signature>"
+````
+
+A first-in-series authorization can also be made using the `applepay`
+payment method; subsequent-in-series authorizations, however,
+must be made using the `card` payment method using the card details of the
+referenced previous-in-series authorization.
+
+Any first-in-series authorization must be made with strong customer
+authentication (SCA) regardless of the authorization amount.
```
tjconcept, line 398: And if not a "one-leg out"?

source/index.html.md
```diff
@@ -540,26 +545,115 @@ Exactly one payment method must be used.
...
+    <div class="type">Conditional. Cannot be present if <code>series[previous]</code> is present.</div>
+  </dd>
+
+  <dt>
+    series[previous][id]
+    <span class="type">[:UUIDv4:]</span>
+  </dt>
+  <dd>
+    The Clearhaus authorization ID as a reference to the latest approved
+    authorization in the series.
```
tjconcept, line 642: Must it be *the* latest or could it be any in the series?

tjconcept (commented):
Sorry for the late comments. The many past diverging PRs scared me off.

source/index.html.md
ct-clearhaus: Should be *the* latest. Another previous-in-series may work, but we aim to increase our garbage collection of old transactions — but we'll ensure the latest is available as long as the series can be continued 🙂

source/index.html.md
ct-clearhaus: https://github.com/clearhaus/gateway-api-docs/pull/118

PR #113, opened 2020-10-21
Add minimalistic MPO API changes

ad-clearhaus:
Adding `cryptogram` and `eci` as optional parameters. Retaining MPO's ability to use non-token based PAN and 3DS.
This ensures backwards compatibility, MPO 3DSecure Fallback and Mastercard support.

For a more full fleshed integration we can opt for receiving the entire `tokenCallback` object from VTS. According to MPO docs, this is what the PSP's are expected to receive from VTS https://github.com/MobilePayDev/MobilePay-Online/pull/25/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R97-R134

This will have to be extended once we figure out what MPO plans to do with Mastercard.

This PR should be read en conjunction with 
* https://github.com/clearhaus/klovedal/pull/480
* https://github.com/clearhaus/authorizer/pull/173

### DIFF
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -687,7 +687,7 @@ object][ApplePay-PaymentToken] for more information.
   </dt>
   <dd>
     Primary account number of card to charge.<br />
-    If <code>token_callback</code> parameter is included, the <code>pan</code> must be a Token PAN.
+    If <code>payment_token</code> parameter is included, the <code>pan</code> must be a Token PAN.
   </dd>
   <dt>mobilepayonline[expire_month]
     <span class="type">[0-9]{2}</span>
@@ -701,7 +701,7 @@ object][ApplePay-PaymentToken] for more information.
   <dd>
     Expiry year of card to charge.
   </dd>
-  <dt>mobilepayonline[token_callback]
+  <dt>mobilepayonline[payment_token]
     <span class="type">[:json:]</span>
   </dt>
   <dd>
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -706,7 +706,7 @@ object][ApplePay-PaymentToken] for more information.
   </dt>
   <dd>
     Full tokenCallback response serialized as JSON, supplied as a string.
-    Required on token based authentication.
+    Required for token-based authentication.
     <br>
     Example: <code>{"paymentId":"string","tokenData":{"cryptogramInfo":{...},...},...}</code>
     <div class="type">Optional</div>
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -686,7 +686,8 @@ object][ApplePay-PaymentToken] for more information.
     <span class="type">[0-9]{12,19}</span>
   </dt>
   <dd>
-    Primary account number of card to charge.
+    Primary account number of card to charge.<br />
+    If <code>cryptogram</code> parameter is included, the <code>pan</code> must be a Token PAN.
   </dd>
   <dt>mobilepayonline[expire_month]
     <span class="type">[0-9]{2}</span>
@@ -699,6 +700,23 @@ object][ApplePay-PaymentToken] for more information.
   </dt>
   <dd>
     Expiry year of card to charge.
+  </dd>
+    <dt>mobilepayonline[cryptogram]
+    <span class="type">[:base64:]{19,21}</span>
+  </dt>
+  <dd>
+    Cryptogram extracted from from MobilePay Online Token Data. Required on
+    token based authentication
+    <div class="type">Optional</div>
+  </dd>
+  </dd>
+    <dt>mobilepayonline[eci]
+    <span class="type">[0-9]{2}</span>
+  </dt>
+  <dd>
+    Eci (Electronic Commerce Indicator) extracted from MobilePay Online Token
+    Data. Only allowed on token based authentication.
+    <div class="type">Optional</div>
   </dd>
   <dt>mobilepayonline[phone_number]
     <span class="type">[\x20-\x7E]{1,15}</span>
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -687,7 +687,7 @@ object][ApplePay-PaymentToken] for more information.
   </dt>
   <dd>
     Primary account number of card to charge.<br />
-    If <code>cryptogram</code> parameter is included, the <code>pan</code> must be a Token PAN.
+    If <code>token_callback</code> parameter is included, the <code>pan</code> must be a Token PAN.
   </dd>
   <dt>mobilepayonline[expire_month]
     <span class="type">[0-9]{2}</span>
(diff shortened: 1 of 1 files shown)


source/index.html.md
```diff
@@ -686,7 +686,8 @@ object][ApplePay-PaymentToken] for more information.
     <span class="type">[0-9]{12,19}</span>
   </dt>
   <dd>
-    Primary account number of card to charge.
+    Primary account number of card to charge.<br />
+    If <code>cryptogram</code> parameter is included, the <code>pan</code> must be a Token PAN.
```
mt-clearhaus, line 690:
```suggestion
    Primary account number (PAN) of card to charge.<br />
    If <code>cryptogram</code> parameter is included, the PAN must be a Token PAN.
```

I notice we use PAN in many places without having introduced it, and we also stubbornly write "primary account number" in a few places. Perhaps we ought to introduce PAN earlier and then only use PAN after that point.

source/index.html.md
```diff
@@ -699,6 +700,23 @@ object][ApplePay-PaymentToken] for more information.
   </dt>
   <dd>
     Expiry year of card to charge.
+  </dd>
+    <dt>mobilepayonline[cryptogram]
+    <span class="type">[:base64:]{19,21}</span>
+  </dt>
+  <dd>
+    Cryptogram extracted from from MobilePay Online Token Data. Required on
```
mt-clearhaus, line 708:
```suggestion
    Cryptogram extracted from from MobilePay Online token data. Required on
```

Or is there a good reason to capitalize token data? We are not referencing a specification or similar here, from where the reader should know about the term.

source/index.html.md
```diff
@@ -699,6 +700,23 @@ object][ApplePay-PaymentToken] for more information.
...
+    token based authentication
```
mt-clearhaus, line 709:
```suggestion
    token-based authentication
```

mt-clearhaus, line 709:
```suggestion
    token-based authentication.
```

source/index.html.md
```diff
@@ -699,6 +700,23 @@ object][ApplePay-PaymentToken] for more information.
...
+    <div class="type">Optional</div>
+  </dd>
+  </dd>
+    <dt>mobilepayonline[eci]
+    <span class="type">[0-9]{2}</span>
+  </dt>
+  <dd>
+    Eci (Electronic Commerce Indicator) extracted from MobilePay Online Token
```
mt-clearhaus, line 717:
```suggestion
    Electronic Commerce Indicator (ECI) extracted from the MobilePay Online token.
```

Or perhaps "token data" like in line 708?

source/index.html.md
```diff
@@ -699,6 +700,23 @@ object][ApplePay-PaymentToken] for more information.
...
+    Data. Only allowed on token based authentication.
```
mt-clearhaus, line 718:
```suggestion
    Data. Only allowed on token-based authentication.
```

source/index.html.md
mt-clearhaus: Ah. Data is on the next line :smile:

source/index.html.md
mt-clearhaus, line 718:
```suggestion
    data. Only allowed on token-based authentication.
```

source/index.html.md
ad-clearhaus: I guess not. :)

ct-clearhaus:
> For a more full fleshed integration we can opt for receiving the entire `tokenCallback` object from VTS

I will vote for the "full 🥩" version 🙂

source/index.html.md
```diff
@@ -699,6 +700,23 @@ object][ApplePay-PaymentToken] for more information.
   </dt>
   <dd>
     Expiry year of card to charge.
+  </dd>
+    <dt>mobilepayonline[cryptogram]
+    <span class="type">[:base64:]{19,21}</span>
```
ct-clearhaus, line 705: A tAV is usually 28 chars after Base64. Do you have a reference for this?

source/index.html.md
```diff
@@ -699,6 +700,23 @@ object][ApplePay-PaymentToken] for more information.
...
+    <div class="type">Optional</div>
+  </dd>
+  </dd>
+    <dt>mobilepayonline[eci]
+    <span class="type">[0-9]{2}</span>
+  </dt>
+  <dd>
+    Electronic Commerce Indicator (ECI) extracted from the MobilePay Online token.
+    data. Only allowed on token-based authentication.
+    <div class="type">Optional</div>
```
ct-clearhaus, line 719: I suspect it's required for Visa?

source/index.html.md
ad-clearhaus: No. I was high (apparently). I have used 28 during my testing also, and the code requires 28.

ad-clearhaus, line 705:
```suggestion
    <span class="type">[:base64:]{28}</span>
```

source/index.html.md
ad-clearhaus: This is the classic very sparse API docs phenomenon at work. If we where to also receive Mastercard in the minimalistic way, we would only receive a `cryptogram` and not `eci`. And instead of getting bugged down in _if this then that, unless a, b, c but only if also d_ I have left it optional and the gateway will be the one responding with _"ECI missing on Visa, bla bla bla"_. 

This is checked by `check_threed_secure_data_for_visa()`
https://github.com/clearhaus/klovedal/blob/master/lib/klovedal/filter/payment_method/token_common.rb#L47-L49

source/index.html.md
```diff
@@ -686,7 +686,8 @@ object][ApplePay-PaymentToken] for more information.
     <span class="type">[0-9]{12,19}</span>
   </dt>
   <dd>
-    Primary account number of card to charge.
+    Primary account number of card to charge.<br />
+    If <code>token_callback</code> parameter is included, the <code>pan</code> must be a Token PAN.
```
mt-clearhaus, line 690: Is "token callback" a phrase used by both Visa and Mastercard? I think the current state of this PR suggests to the reader that it is.

source/index.html.md
```diff
@@ -700,6 +701,16 @@ object][ApplePay-PaymentToken] for more information.
   <dd>
     Expiry year of card to charge.
   </dd>
+  <dt>mobilepayonline[token_callback]
+    <span class="type">[:json:]</span>
+  </dt>
+  <dd>
+    Full tokenCallback response serialized as JSON, supplied as a string.
+    Required on token based authentication.
```
mt-clearhaus, line 709:
```suggestion
    Required on token-based authorization.
```

mt-clearhaus, line 709:
```suggestion
    Required for token-based authorization.
```

source/index.html.md
ad-clearhaus: I would say that `tokenCallback` is the closest MPO comes to giving it a name. It is the entire _thing_ returned from the tokenization service. 

https://github.com/MobilePayDev/MobilePay-Online/pull/25/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R101https://github.com/MobilePayDev/MobilePay-Online/pull/25/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R101

> When you recieve the tokenCallback, you´ll find a cardIssuedInCountryCode (possible values DK, FI) you can use for your Acquirer routing logic. And a Visa Token Service (VTS) service response like this: <br />

I have purposely used MPO terminology to avoid having to pick between Visa or Mastercard terminology.

But the naming link in flimsy, and a better option would be preferable. Golf away! :golfing:

mt-clearhaus: Fair :slightly_smiling_face: It seems, however, `tokenCallback` is only mentioned in the "Delegated Authentication for all Visa cards" section, not in the "Delegated Authentication for all Mastercards" section. It will be interesting to see what happens w.r.t. Mastercard support. It is likely a favorable bet that it will also involve a `tokenCallback`, as they're also aiming for tokenization with Mastercard and state that both tokenization schemes are based on EMVCo.

ad-clearhaus: Let us settle on a parameter name: Currently: `token_callback`. Solely based on

> When you recieve the tokenCallback, you´ll find a cardIssuedInCountryCode (possible values DK, FI) you can use for your Acquirer routing logic.

Which is not good.
Just `token` is a bit to generic and could be interpreted as just the `tokenData` part of the data received on tokenCallback. Alternatively `token_callback_data`.

Finally we can just go with `payment_token` just like `applepay`.

ct-clearhaus: > Finally we can just go with `payment_token` just like `applepay`.

I like the strategy and think `payment_token` is a decent name. 👍

mt-clearhaus (approved):
:+1:

PR #117, opened 2020-12-21
Replace ' ' with ' '. Hex 00a0 vs. Hex 20

ad-clearhaus:
A non-ASCII char had sneaked in:

Screenshot from 2020-12-21 10-13-23


mt-clearhaus (approved):
Nice catch! :+1:

PR #119, opened 2021-01-29
Add vipps

ad-clearhaus:
Adds Vipps as new `payment_method`

For Vipps API descirption see: https://github.com/vippsas/vipps-psp-api/blob/master/vipps-psp-api.md#emvco-token-processing

Some digging has been done in https://github.com/clearhaus/issues-company/issues/66


ct-clearhaus (approved):
👍

tjconcept:
Wait, what is happening? Is it moving out of beta anyway? 😀

ct-clearhaus:
> Wait, what is happening? Is it moving out of beta anyway? 😀

It seems that I am publicly leaking some of the news coming in the Q4 partner bulletin that could be circulated in a not too distant future 😄 ... so let me just continue that adventure 😏

We've seen a bit of Vipps transactions (100-500 daily) over the last rolling quarter with a decent approval rate (daily: min=~90%, avg=~96%, max=~99%), so we are confident that it's not completely broken and are ready to see more traffic 🙂

PR #122, opened 2021-02-11
Expose currency on account resource

ct-clearhaus:


source/index.html.md
```diff
@@ -1239,6 +1239,12 @@ GET https://gateway.clearhaus.com/account
   <dd>
     ISO 3166-1 2-letter country code for merchant company.
   </dd>
+  <dt>country
```
dvaeversted, line 1242: Shouldn't this be "currency" instead ?

tjconcept:
I'm curious what use case motivated this?

source/index.html.md
ct-clearhaus: Certainly!
```suggestion
  <dt>currency
```

ct-clearhaus:
> I'm curious what use case motivated this?

One merchant can have multiple accounts where the only difference is the funding currency. This allows to easily distinguish which of such accounts a given API key is for. Could be achieved via naming or alike, but `currency` is consistent and always correct; a naming convention is more relaxed and can be mixed up.

kse-clearhaus (approved):

PR #125, opened 2021-03-16
Return series

ct-clearhaus:


ct-clearhaus:
While investigating a case with @absrd I came to think if we should amend the response with "first or subsequent".

## Should we?

It's kinda 🙃 because ... previously with the now deprecated `recurring=(true|false)` it was a bit hard for the client to know if it was first-in-series or subsequent-in-series, but now with `series[type]` and `series[previous][]` the client is able to decide if it should be first-in-series or subsequent-in-series. So ... with the deprecated API it's relevant, with the new API it's not very relevant because it's already decided explicitly by the client.

WDYT?

## How?

My immediate suggestion would be

```diff
  {
      "id": "1b722683-92ad-4c6b-85da-e119d550670d",
      "status": { "code": 20000 },
      "series": {
          "type": "recurring",
~         "tid": "481048839682954",
+         "first": true,
+         "subsequent": false
      }
  }
```

🤷‍♂️

ct-clearhaus:
I turned down the idea to show "first or subsequent".

ct-clearhaus:
For Mastercard, an "exemption" parameter is responded:
```json
"series": {
    "type": "recurring",
    "tid": "MCSA2C4E60527  ",
    "exemption": "variable_amount_series"
}
```

ct-clearhaus:
The CircleCI job to deploy this failed 😞

ct-clearhaus:
Fixed in https://github.com/clearhaus/gateway-api-docs/pull/126 (courtesy of @Lassejoe) 🥳

PR #127, opened 2021-06-02
Introduce unscheduled (UCOF) series

ct-clearhaus:
This is building upon the design and decisions in https://github.com/clearhaus/gateway-api-docs/pull/111 and is thus heavily inspired by 3d0c2e5.

To be fixed:

* [x] When migrating a series to us from another acquirer (using the raw scheme TID etc.), we need to know what type it is.
  * How to get to know this?
  * By supplying ...
    * (a) both `series[type]` and `series[previous][$scheme]`?
    * (b) adding `series[previous][$scheme][type]` and thus supplying it "next to" `series[previous][$scheme][tid]`?
    * (c) adding `series[previous][type]` and thus supplying it with `series[previuos][$scheme][tid]`?
  * Required or optional? (If optional, what should be the default?)

### DIFF
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -340,8 +340,8 @@ Clearhaus supports two types of subscription billing:
 
 * Recurring: Transactions processed at agreed, predetermined, regular intervals not
     exceeding 1 year; e.g. a monthly subscription for a magazine.
-* Unscheduled (UCOF, Unscheduled Credential on File): Transactions that does not
-    occur on predetermined, regular intervals; e.g. a car sharing subscription
+* Unscheduled (UCOF, Unscheduled Credential on File): Transactions not occurring
+    on predetermined, regular intervals; e.g. a car sharing subscription
     billed weekly but only for weeks when the service is used.
 
 For both types, there may be an agreed end of the series and the amount may be
@@ -643,7 +643,7 @@ Exactly one payment method must be used.
     <br />
     <code>unscheduled</code>: A series of transactions where the cardholder has
     explicitly agreed that the merchant may repeatedly charge the cardholder at
-    unknown times, e.g. based on cardholder usage.
+    non-predetermined times, e.g. based on cardholder usage.
 
     <div class="type">Conditional. Cannot be present if <code>series[previous]</code> is present.</div>
   </dd>
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -336,16 +336,22 @@ parameter `card[name]`.
 
 ## Series of transactions
 
-Clearhaus supports the recurring type of subscription billing.
+Clearhaus supports two types of subscription billing:
 
-There may be an agreed end of the series. The amount may be varying. See the
-partner guideline for more details.
+* Recurring: Transactions processed at agreed predetermined, regular intervals not
+    exceeding 1 year; e.g. a monthly subscription for a magazine.
+* Unscheduled (UCOF, Unscheduled Credential on File): Transactions that does not
+    occur on predetermined, regular intervals; e.g. a car sharing subscription
+    billed weekly but only for weeks when the service is used.
+
+For both types, there may be an agreed end of the series and the amount may be
+varying. See the partner guideline for more details.
 
 ### Repeatedly reserve money
 
-A first-in-series recurring payment is created by making an authorization and
-marking it as a `recurring` series. As an example, a first-in-series recurring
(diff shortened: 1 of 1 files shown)


source/index.html.md
```diff
@@ -336,16 +336,22 @@ parameter `card[name]`.
 
 ## Series of transactions
 
-Clearhaus supports the recurring type of subscription billing.
+Clearhaus supports two types of subscription billing:
 
-There may be an agreed end of the series. The amount may be varying. See the
-partner guideline for more details.
+* Recurring: Transactions processed at agreed predetermined, regular intervals not
```
mt-clearhaus, line 341:
```suggestion
* Recurring: Transactions processed at agreed, predetermined, regular intervals not
```

source/index.html.md
```diff
@@ -336,16 +336,22 @@ parameter `card[name]`.
...
+    exceeding 1 year; e.g. a monthly subscription for a magazine.
+* Unscheduled (UCOF, Unscheduled Credential on File): Transactions that does not
+    occur on predetermined, regular intervals; e.g. a car sharing subscription
```
mt-clearhaus, line 344:
```suggestion
* Unscheduled (UCOF, Unscheduled Credential on File): Transactions not occurring
    on predetermined, regular intervals; e.g. a car sharing subscription
```

source/index.html.md
```diff
@@ -626,14 +632,18 @@ Exactly one payment method must be used.
 
   <dt>
     series[type]
-    <span class="type"><code>recurring</code></span>
+    <span class="type">(<code>recurring</code>|<code>unscheduled</code>)</span>
```
mt-clearhaus, line 635:
```suggestion
    <span class="type"><code>(recurring|unscheduled)</code></span>
```

source/index.html.md
```diff
@@ -626,14 +632,18 @@ Exactly one payment method must be used.
...
   <dd>
     The type of series.
     <br />
     <code>recurring</code>: A series of transactions where the cardholder has
     explicitly agreed that the merchant may repeatedly charge the cardholder at
     regular, predetermined intervals that may not exceed 1 year.
+    <br />
+    <code>unscheduled</code>: A series of transactions where the cardholder has
+    explicitly agreed that the merchant may repeatedly charge the cardholder at
+    unknown times, e.g. based on cardholder usage.
```
mt-clearhaus, line 646:
```suggestion
    non-predetermined times, e.g. based on cardholder usage.
```

source/index.html.md
ct-clearhaus: Is that how it's formatted elsewhere? (I thought I followed the standard of how it's done elsewhere, but I'm probably wrong 🙂)

mt-clearhaus: You're right. It's formatted with blended code and non-code in other places. Let's just keep it that way. I don't understand, however, why someone chose to not include the parentheses and the pipe inside `<code>` in the first place.

mt-clearhaus: Looking at it rendered, I acknowledge that it might have some value that the possible keywords (here "recurring" and "unscheduled") are emphasized rather than being grey like with the other regular expressions. Strictly speaking, we are (ab)using regular expression syntax, so consistency dictates 100% greyness, however, it has to be weighed against the value of emphasized values, and I tend to say the latter "wins".

ct-clearhaus: I agree completely, but I think I'll just follow whatever is apparently the default to avoid scope creep 🙂

dvaeversted:
Now that you are mentioning the case of migrating it from 1 merchant agreement to another within Clearhaus.
And the case where a merchant is to be migrated away from Clearhaus.

Then it would be nice to always expose the trace id to the PSP, so we are able to save it together with the other data we have on a COF transaction. Even if we in normal day-to-day are not supposed to use it for anything :)

ct-clearhaus:
> Then it would be nice to always expose the trace id to the PSP, so we are able to save it together with the other data we have on a COF transaction. Even if we in normal day-to-day are not supposed to use it for anything :)

https://github.com/clearhaus/gateway-api-docs/pull/125/files#diff-953ec4c8cae7c7113d7ae3d7c5ec008a0d8a750e363b5473bfb3db7debe0cf20R373 ✅ 🙂

ct-clearhaus:
This change is available on `gateway.test.clearhaus.com` 🙂

PR #129, opened 2021-06-14
Add mail and telephone as payment methods

ct-clearhaus:

PR #131, opened 2021-06-17
Add moto as new payment method

mt-clearhaus:
Alternative to https://github.com/clearhaus/gateway-api-docs/pull/129. I'm a bit more fond of this, as the lines here don't look as much copy-pasty (the exact same parameters and notes for mail and telephone).

### DIFF
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -539,7 +539,7 @@ POST https://gateway.clearhaus.com/authorizations
 ````
 
 Authorizations can be created using different payment methods:
-`card`, `applepay`, `mobilepayonline`, `mail`, `telephone`.
+`card`, `applepay`, `mobilepayonline`, `moto`.
 Exactly one payment method must be used.
 
 #### Parameters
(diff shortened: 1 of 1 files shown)


kse-clearhaus:
Note that we plan on allowing CSC for telephone orders.

mt-clearhaus:
It seems our compliance folks might at some point conclude that CSC shall never be sent upstream for either mail or telephone, at least for one of the schemes. The outcome of their investigation might be the death blow for this combined approach.

AlexanderIversen:
> The outcome of their investigation might be the death blow for this combined approach.

I'd say that the option to get liability shift for Visa Telephone Orders when CVV2 is included and the issuer approves the transaction also could be the death blow of this option. If we asked the Risk team I believe they would vote for an option where we can separate MO and TO since there exist scenarios of liability shift. 

Whether any of the current merchants and supposedly also coming merchant support such options is unknown. 

Remember I am no pro at reading HTML. :)

mt-clearhaus:
@AlexanderIversen Note that you **can** separate MO and TO in this approach, as `moto[type]` is not optional.

mt-clearhaus:
The quote "Due to Visa not distinguishing between MO and TO in authorization including/excluding CVV2 in authorization may actually indicate to Visa whether the transaction is MO or TO, since the requirement only applies to one of the scenarios (TO). We only have the option to populate BM 25 = 08 (MOTO) for Visa." by @AlexanderIversen in https://clearhaus.slack.com/archives/C024FEWDBEK/p1623933187004800?thread_ts=1623868130.030800&cid=C024FEWDBEK has made me doubt the combined approach in this PR. If he is right, which I have no reason to doubt, I think it follows that we don't want to let clients send us CSC for a Visa MO. And IMO we only want to let clients send us a parameter if we are going to send it upstream in some way.

tjconcept:
> If he is right, which I have no reason to doubt, I think it follows that we don't want to let clients send us CSC for a Visa MO. And IMO we only want to let clients send us a parameter if we are going to send it upstream in some way.

Why? Is it something about giving the impression that it's used while it's not? Isn't it sort-of already the case and hence the `{present, matches}` return value? Maybe it could be extended with `ignored` or similar?

Just tossing in random thoughts from the sideline, lacking background, ignore if I'm obviously ignorant.

kse-clearhaus:
> Why? Is it something about giving the impression that it's used while it's not? Isn't it sort-of already the case and hence the `{present, matches}` return value? Maybe it could be extended with `ignored` or similar?

I wish we could expose that, but I don't think we'd be able to conclusively say in "enough" cases that it would be ignored or not.

ct-clearhaus:
> I don't think we'd be able to conclusively say in "enough" cases that it would be ignored or not.

In the response we can for sure tell if we ignored the supplied CSC though.

kse-clearhaus:
> > I don't think we'd be able to conclusively say in "enough" cases that it would be ignored or not.
> 
> In the response we can for sure tell if we ignored the supplied CSC though.

Absolutely, we already do implicitly, if a CSC is supplied but not sent upstream, we return
```
  "csc": {
    "matches": false,
    "present": false
  }
```

mt-clearhaus:
Based on today's MOTO API meeting, I have removed `moto[csc]` and `moto[type]`. We will launch without those and only add them if they become necessary. It follows that we will only process MO, not TO. The situation is that the client will not tell us if a transaction is MO or TO, and it is our decision to process the transaction as MO.

ct-clearhaus:
The feature is now available on `gateway.test.clearhaus.com` 🙂

ct-clearhaus (approved):

PR #133, opened 2021-07-14
Bugfix Mastercard exemption values

ct-clearhaus:
The names are:

DE48: Additional Data
DE48SE22: Multi-Purpose Merchant Indicator
DE48SE22SF01: Low-Risk Merchant Indicator

However, I think these are of no value in the documentation, so I decided to leave it out; at least it's better than indicate a naming that is wrong.

Courtesy of @absrd 🙇‍♂️


AlexanderIversen (commented):
LGTM

PR #135, opened 2021-09-10
Clarify necessity of COF=store before MIT

ct-clearhaus:
I *think* this is a clarification, and I *think* it's clear enough, but please be liberal with critics and suggestions 🙂

### DIFF
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -598,7 +598,7 @@ Exactly one payment method must be used.
     This is regardless of whether a stored payment credential is being used.
     <br />
     For compliance reasons there should be a previous approved transaction (for
-    the combination of cardholder and merchant) where
+    the combination of card and merchant) where
     <code>credential_on_file=store</code> before <code>initiator</code> may be
     <code>merchant</code>.
     <br />
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -598,8 +598,9 @@ Exactly one payment method must be used.
     This is regardless of whether a stored payment credential is being used.
     <br />
     For compliance reasons there should be a previous approved transaction (for
-    the cardholder and the merchant) marked with <code>storing</code> before
-    <code>initiator</code> may be <code>merchant</code>.
+    the combination of cardholder and merchant) where
+    <code>credential_on_file=store</code> before <code>initiator</code> may be
+    <code>merchant</code>.
     <br />
     Default:
     <ul>


ct-clearhaus:
For reference: https://clearhaus.slack.com/archives/CT5ETNF33/p1631197619005400

source/index.html.md
```diff
@@ -598,8 +598,9 @@ Exactly one payment method must be used.
     This is regardless of whether a stored payment credential is being used.
     <br />
     For compliance reasons there should be a previous approved transaction (for
-    the cardholder and the merchant) marked with <code>storing</code> before
-    <code>initiator</code> may be <code>merchant</code>.
+    the combination of cardholder and merchant) where
```
mt-clearhaus, line 601:
```suggestion
    the combination of card and merchant) where
```

Strictly speaking, it's the card, not the cardholder, because a cardholder might have more than a single card :man_shrugging: Perhaps we could or should write PAN instead, but I leans towards just card.

Other than that I think the clarification makes sense and is clear.

mt-clearhaus:
When we're around here anyway, let me highlight the following sentence:

> An authorization is initiated by the cardholder if the cardholder decided the transaction should be created.

Can we make this less vague? In my view, subsequent-in-series recurring authorizations for e.g. a magazine subscription is also something that the the cardholder decided to "order" when he/she signed up in the first place. And we don't mean that here, because those subsequent-in-series authorizations are MITs. Perhaps "cardholder decided" ought to somehow be replaced with "cardholder present", however, preferably something else I think, as "cardholder present" has very specific meanings with schemes and might be more POS than ecommerce.

ct-clearhaus:
> Can we make this less vague?

Uhm, I guess this is more or less stolen from Visa and Mastercard which says, respectively:

> Visa "Improving Authorization Management for Transactions with Stored Credentials", section "Stored Credential Terminology", p. 6:
> Cardholder-initiated Transaction (CIT): A cardholder-initiated transaction is any transaction where the cardholder is actively participating in the transaction. This can be either at a terminal in-store or through a checkout experience online, or with a stored credential.

> Mastercard Authentication Guide for Europe, v 1.2, p. 43:
> Merchant-Initiated Transactions are payments initiated by the Merchant without the interaction of the payer. They are characterized by a lack of involvement of the payer in triggering each individual payment.

I'm not quite sure how to tighten the wording.

mt-clearhaus:
Aah, I see :+1:  What we have isn't that bad then. However, perhaps a tiny improvement could be to replace "decided the transaction should be created" with "is actively participating in the transaction"? Well, it _is_ tiny, and it's not really in scope for this PR :man_shrugging: Feel free to merge without further ado.

mt-clearhaus (approved):

PR #137, opened 2021-09-22
Raw Apple Pay

mt-clearhaus:
Issue: https://github.com/clearhaus/issues-pci/issues/2243

Document the creation of an Apple Pay authorization from raw values from a payment token.

### DIFF
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -793,21 +793,6 @@ object][ApplePay-PaymentToken] for more information.
   </dd>
 </dl>
 
-<p class="alert alert-info">
-  <b>Notice:</b> Signing is required to use the <code>applepay</code> payment
-  method.
-  <br />
-  <b>Notice:</b> An authorization made with <code>applepay</code> is
-  strongly authenticated (SCA in PSD2).
-  <br />
-  <b>Notice:</b> An authorization made with <code>applepay</code> may be 3-D
-  Secured to some degree or not at all; this is indicated by the
-  <code>eciIndicator</code> of the <code>applepay[payment_token]</code>.
-  <br />
-  <b>Notice:</b> An authorization made with <code>applepay</code> cannot be a
-  subsequent-in-series authorization.
-</p>
-
 Additionally, an Apple Pay authorization can be created using raw values from
 a payment token:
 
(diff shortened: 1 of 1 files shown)


source/index.html.md
```diff
@@ -808,6 +808,42 @@ object][ApplePay-PaymentToken] for more information.
...
+    Expiry month of card to charge.
+  </dd>
+  <dt>applepay[raw][expire_year]
+    <span class="type">20[0-9]{2}</span>
+  </dt>
+  <dd>
+    Expiry year of card to charge.
+  </dd>
+  <dt>applepay[raw][cryptogram]
+    <span class="type">[:base64:]{28}</span>
```
mt-clearhaus, line 834: This is the first time we specify a length requirement on a base64 value in the documentation. I don't know if this way is acceptable :slightly_smiling_face:

ct-clearhaus: We do this [several](https://github.com/clearhaus/gateway-api-docs/blame/d7203f8be6bc60d82ff0c536ddba27fb5736a06d/source/index.html.md#L557) [other](https://github.com/clearhaus/gateway-api-docs/blame/d7203f8be6bc60d82ff0c536ddba27fb5736a06d/source/index.html.md#L564) [places](https://github.com/clearhaus/gateway-api-docs/blame/d7203f8be6bc60d82ff0c536ddba27fb5736a06d/source/index.html.md#L626) so I think it's perfectly fine 🙂

source/index.html.md
```diff
@@ -808,6 +808,42 @@ object][ApplePay-PaymentToken] for more information.
   subsequent-in-series authorization.
 </p>
 
+Additionally, an Apple Pay authorization can be created using raw values from
+a payment token:
```
ct-clearhaus, line 812: Does it belong here to mention that the TP should reach out to ensure that they're allowed to use `applepay[raw]`?

What about listing the requirements the the TP should enforce? (Check for signature, for age, and other stuff mentioned in https://developer.apple.com/library/archive/documentation/PassKit/Reference/PaymentTokenJSON/PaymentTokenJSON.html)

source/index.html.md
```diff
@@ -808,6 +808,42 @@ object][ApplePay-PaymentToken] for more information.
...
+    Expiry year of card to charge.
+  </dd>
+  <dt>applepay[raw][cryptogram]
+    <span class="type">[:base64:]{28}</span>
+  </dt>
+  <dd>
+    Online payment cryptogram.
+  </dd>
+  <dt>applepay[raw][eci]
+    <span class="type">[0-9]{1,2}</span>
```
ct-clearhaus, line 840: I stumble upon the variable length. Usually ECI is always 2, but in the Apple Pay tokens we receive it's always 1 (just hunted it down to be sure that we log it "very raw"); and [the spec](https://developer.apple.com/library/archive/documentation/PassKit/Reference/PaymentTokenJSON/PaymentTokenJSON.html) doesn't mention anything about length 🙁

* `[1-9][0-9]?`? Nah. That's likely "too regexp-y" 🙁
* `([1-9]|[1-9][0-9])`? Nah, don't like it.

Uhm, couldn't we just require it to be 2 long? PSPs doing integrations "more directly" like this, they'd likely need to make the ECI be 2 anyway ... WDYT?

source/index.html.md
```diff
@@ -808,6 +808,42 @@ object][ApplePay-PaymentToken] for more information.
...
+  </dt>
+  <dd>
+    Electronic Commerce Indicator.
```
ct-clearhaus, line 843: Maybe mention that it comes from `eciIndicator`?

ct-clearhaus (commented):
Looking good!

source/index.html.md
mt-clearhaus: I actually thought `[0-9]{1,2}` was a good compromise :smile: However, I also like the idea about being "strict". I'll take a look the code.

source/index.html.md
mt-clearhaus: https://github.com/clearhaus/gateway-api-docs/pull/137/commits/1e6b4a3190907b1aff08633f01d01652e45b9002

source/index.html.md
mt-clearhaus: https://github.com/clearhaus/gateway-api-docs/pull/137/commits/be79759f58f5875fa869c569489da563c6e1304f

I moved the alert box down because its notices are also relevant for raw Apple Pay. Both non-raw and raw require that a TP is allowed to use AP.

source/index.html.md
mt-clearhaus: https://github.com/clearhaus/gateway-api-docs/pull/137/commits/c39552d091539cbe66ecceff8400d0f1090962cb

ct-clearhaus:
We have confirmed that this works on production systems for both Visa and Mastercard.

Issue #143, opened 2022-03-09
Discrepancy in documentation


misozask:
Please update the documentation of `amount` in https://docs.gateway.clearhaus.com/#authorizations
10 digits are not allowed, just 9.

![image](https://user-images.githubusercontent.com/32176893/157469174-2e330a4f-9bd2-495e-99f9-0e492a22b028.png)

mt-clearhaus:
@misozask Thank you very much for reporting this. I have created https://github.com/clearhaus/gateway-api-docs/pull/144.

PR #144, opened 2022-03-09
Correct regular expressions for amount

mt-clearhaus:
Fixes https://github.com/clearhaus/gateway-api-docs/issues/143.

### DIFF
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -555,7 +555,7 @@ Exactly one payment method must be used.
 
 <dl class="dl-vertical">
   <dt>amount
-    <span class="type">[0-9]{1,10}</span>
+    <span class="type">(0|[1-9][0-9]{,8})</span>
   </dt>
   <dd>
     Amount in minor units of given currency (e.g. cents if in Euro).
@@ -1283,7 +1283,7 @@ POST https://gateway.clearhaus.com/authorizations/:id/captures
 
 <dl class="dl-vertical">
   <dt>amount
-    <span class="type">[1-9][0-9]{1,9}</span>
+    <span class="type">[1-9][0-9]{,8}</span>
   </dt>
   <dd>
     Amount in minor units of given currency (e.g. cents if in Euro).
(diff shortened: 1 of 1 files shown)


ct-clearhaus (approved):
👍

misozask:
Are you sure that `(0|[1-9][0-9]{,8})` is correct regex?
I assume it should rather be `^(0|[1-9][0-9]{0,8})$`

mt-clearhaus:
`^` and `$` are implied in the regexes mentioned in the documentation. It is optional but valid to leave out `0` in a quantifier.

misozask:
`{,8}` is valid? Doesn't look like as 
> {,8} matches the characters {,8} literally (case sensitive)
 https://regex101.com/r/GV0qpX/1

misozask:
> ^ and $ are implied in the regexes mentioned in the documentation

Somehow I can't find it ...

mt-clearhaus:
For most applications if you ask them to expect Extended Regular Expressions (EREs), opposed to Basic Regular Expressions (BREs), they will interpret `{a,b}` as a quantifier, not as the five literal characters. Many applications accept `a` or `b` being left out. With `a` left out this means "the preceding element or subexpression must occur at most `b` times.", i.e. no lower bound. With `b` left out this means "the preceding element or subexpression must occur at least `a` times.", i.e. no upper bound. It appears various applications and help sites do not fully agree on a strict distinction between BRE and ERE and whether or not bounds may be left out. There does not appear to be any widespread consensus.

The expressions used in our documention are deliberately intended to be understood by a human rather than an application. We also employ home-made expressions, e.g. `[:UUIDv4:]` for `series[previous][id]`. This is also why we do not think it is necessary to include `^` and `$` in all expressions—or mention explicitly that we have made that choice.

misozask:
Yes, I've figured out that some regex flavors supports also `{,b}`. But IMHO most common ones does not.
https://www.regular-expressions.info/refrepeat.html

So, up to you if you want to make your clients wonder about it (when they blindly copy it, as I did), or rather use syntax recognized in all flavors ...
Just my two cents ...

Thanks anyway!

PR #145, opened 2022-04-04
Update to buster, ruby 2.5.5 and puma 4.3.12

brorbw:
This is a shallow poke since something breaks the website and I have not looked into the details.

Left is before updating right is after
Screenshot 2022-04-04 at 16 37 46


Dockerfile
```diff
@@ -1,7 +1,7 @@
-FROM node:stretch
+FROM node:buster
 
 RUN apt-get update \
-    && apt-get install -y ruby2.3 bundler \
+    && apt-get install -y ruby2.5 bundler \
```
kbadk, line 4: Might be a good idea to go straight to 2.7 while you're at it. 2.7 is what we're running in most other images.

brorbw: I updated to 2.5 as an initial test to see how much broke. This is only on a `buster` image also.

brorbw:
Closing due to https://clearhaus.slack.com/archives/C3LG75WE9/p1649421708553339

PR #146, opened 2022-05-11
Add debits

mt-clearhaus:
Add the `/debits` resource which is an umbrella for transferring money from a cardholder's bank account in one step.

Issue: https://github.com/clearhaus/issues-pci/issues/2777


tjconcept:
Interesting reading. Do you have thoughts about idempotence in relation to this? What happens if the connection breaks between issuing a request for a debit and the response being booked?

mt-clearhaus:
Rendered version of the changes at this stage: [rendered-20220511.pdf](https://github.com/clearhaus/gateway-api-docs/files/8670108/rendered-20220511.pdf)
Regarding the colored boxes to the right, green means "added" whereas blue means "changed".

mt-clearhaus:
@tjconcept I do not have any answers in relation to idempotence at this stage but thank you for bringing up the topic. I will make sure the topic is given some thought in the near future.

mt-clearhaus:
@tjconcept We are aware that the introduction of a one-step debit concept will likely increase the need for idempotency, at least for the new resource. It will not have the advantage of a two-phase commit as you mention. In order to avoid scope creep, however, idempotency will not be included here or in the underlying one-step debit project.

source/index.html.md
```diff
@@ -1491,10 +2238,14 @@ POST https://gateway.clearhaus.com/credits
...
     <div class="type">Optional</div>
   </dd>
-  <dt>card[name]
+
+  <!-- deprecated -->
+  <dt><strike>card[name]</strike>
     <span class="type">[A-Za-z0-9 ]{1,30}</span>
   </dt>
   <dd>
+    Deprecated! Please use <code>recipient[name]</code>.
```
ct-clearhaus, line 2247:
```suggestion
    Deprecated! Parameter will be ignored as of 202?-??-??.
    Please use <code>recipient[name]</code>.
```

source/index.html.md
```diff
@@ -1406,22 +1408,758 @@ can still only capture 5.
...
+      </a>
+    </span>
+  </dt>
+  <dd>
+    A reference to an external object, such as an order number.
+    <div class="type">Optional</div>
+  </dd>
+
+  <dt>
+    series[type]
```
LQkkeN, line 1538: Will the **debit** type support being in-series already? I was under the impression that this feature would be added at some later point.

ct-clearhaus: We'll see how it unfolds — if it's "sufficiently easy" to add it right away, we'll do that; if it ends up being a larger one we may cut the scope initially. However, you're right that it'll very likely come eventually.

mt-clearhaus:
Rendering as of https://github.com/clearhaus/gateway-api-docs/pull/146/commits/683e8cc81bfebd7591204077066e7e043dc97c9c:

Screenshot from 2023-01-05 17-16-38
Screenshot from 2023-01-05 17-18-59
Screenshot from 2023-01-05 17-11-57
Screenshot from 2023-01-05 17-12-19

website/content/gateway/api_reference/resources/debit_refunds/debit_refunds.md
```diff
@@ -0,0 +1,19 @@
...
+
+```shell
+POST https://gateway.clearhaus.com/debits/:id/refunds
+```
+
+##### Parameters
+No parameters are needed to make a new refund transaction.
+
+{{% notice %}}
+**Notice**: A refund of a Visa AFT debit might be declined by Visa or the issuer if the debit age is 24 hours or more, or if settlement has taken place for the debit.
```
mt-clearhaus, line 18: It has been pointed out that "settlement" here is imprecise.

Unfortunately, I have not been able to find any details in the Visa documentation I have read so far.

See https://clearhaus.slack.com/archives/C02V02D1SSC/p1672938054830459?thread_ts=1672935881.101459&cid=C02V02D1SSC.

mt-clearhaus: I have tried to clarify matters in https://github.com/clearhaus/gateway-api-docs/pull/146/commits/ad4f0b84db62fc8b84e58c799719a6c7d7572af3.

cbobach:
Should we state something about the required parameters for cross-border, US, CA?

mt-clearhaus:
> Should we state something about the required parameters for cross-border, US, CA?

So far I have tried to avoid too many details by referring to the partner guideline. I would like input on @clearhaus/transaction-compliance on how to divide the information between this document and the partner guideline (via other channels than here, preferably). I imagine, however, that  we will implement up-front checks in the gateway for e.g. required parameters due to cross-border, so we might as well include the requirements here.

website/content/gateway/api_reference/resources/debit_refunds/debit_refunds.md
```diff
@@ -0,0 +1,19 @@
...
+
+```shell
+POST https://gateway.clearhaus.com/debits/:id/refunds
+```
+
+##### Parameters
+No parameters are needed to make a new refund transaction.
+
+{{% notice %}}
+**Notice**: A refund of a Visa AFT debit might be declined by Visa or the issuer if the debit age is 24 hours or more, or if settlement has taken place for the debit. The transaction gateway does not track debit age or settlement, i.e. it relies on Visa and the issuer to decline late refunds.
```
ct-clearhaus, line 18: Continuing from https://github.com/clearhaus/gateway-api-docs/pull/146#pullrequestreview-1238058763, I suspect "clearing" is what's meant?

```suggestion
**Notice**: A refund of a Visa AFT debit might be declined by Visa or the issuer if the debit age is 24 hours or more, or the debit has cleared. The transaction gateway does not track the clearing of the debit, i.e. it relies on Visa and the issuer to decline late refunds.
```

Do we decline debit refunds when the debit is "too old" — say, the normal ~180 days?

website/content/gateway/api_reference/resources/debits/debits.md
```diff
@@ -0,0 +1,20 @@
...
+date: 2023-01-05T13:37:00+01:00
+anchor: "debits"
+weight: 176
+---
+#### Debits
+To transfer money from a cardholder's bank account in real-time, in one step, you make a new debit resource.
+```shell
+POST https://gateway.clearhaus.com/debits
+```
+Debits are created using the same payment methods and parameters as listed under [Authorizations](#authorizations). Exactly one payment method must be used.
```
ct-clearhaus, line 12:
```suggestion
Debits are created using a payment method ([card](#method-card), [applepay](#method-applepay), [googlepay](#method-googlepay), [mobilepayonline](#method-mobilepayonline), [moto](#method-moto), [token](#method-token), or [vipps](#method-vipps)) as listed under [Authorizations](#authorizations). Exactly one payment method must be used.
```

website/content/gateway/api_reference/resources/debits/debits.md
```diff
@@ -0,0 +1,20 @@
...
+
+For authentication and scheme reference to series see the sections under [Authorizations](#authorizations), respectively. Notice that only 3-D Secure version 2, i.e. `[3dsecure][v2][...]`, is supported for debits, not version 1.
```
ct-clearhaus, line 14:
```suggestion
For authentication, [3-D Secure version 2](#authentication-3dsecurev2) is supported, and also [scheme reference to series](#scheme-reference-to-series) is supported---exactly as it is for [Authorizations](#authorizations). Notice that only 3-D Secure version 1 is not supported.
```

website/content/gateway/api_reference/resources/resources.md
```diff
@@ -5,11 +5,13 @@ anchor: "resources"
 weight: 100
 ---
 ## Resources
-Our API offers six different resources:
+Our API offers eight different resources:
 
 - [Authorizations](#authorizations)
 - [Captures](#captures)
 - [Refunds](#refunds)
```
ct-clearhaus, line 12: Should we rename these to "Capture refunds" to be consistent?

ct-clearhaus (approved):
Awesome!

AlexanderIversen:
> So far I have tried to avoid too many details by referring to the partner guideline. I would like input from https://github.com/orgs/clearhaus/teams/transaction-compliance on how to divide the information between this document and the partner guideline

And yes it is a good question MT and I am not sure. I have asked the question myself the other way around, namely what should go in the partner guideline and what should be in Gateway Docs.

website/content/gateway/api_reference/resources/debit_refunds/debit_refunds.md
mt-clearhaus: That is a good question. It wasn't my plan as I hope to simplify things by letting upstream be the judge :balance_scale: :hammer: 

We might want to implement a check if it's expensive in fees for us to make very late refunds.

@clearhaus/transaction-compliance Do you have any input here?

mt-clearhaus: On another note, we want to limit the amount of time we need to store the PAN. The PAN isn't sent to us in a debit refunds request.

mt-clearhaus: I also tend to say the right word here is "clearing", not "settlement". So I'm inclined to apply your suggestion. I will let the thread hang a bit before doing so.

website/content/gateway/api_reference/resources/resources.md
mt-clearhaus: At first I wasn't in doubt, however, now I am.

Ordinary refunds hang on an authorization, not on a capture. Of course, there is a relationship to an authorization's capture(s), as one or more captures must have taken place before a refund can be made.

Try to taste this sentence from the current Refunds section:

> You can make multiple refunds for an authorization transaction.

Changing "Refunds" to "Capture refunds" will not break things in any deep way, but the heading suggests that one POSTs on `/captures/:id/refunds` which is not the case.

@ct-clearhaus WDYT?

ct-clearhaus: "Authorization refunds" ... I don't like this one because one can in fact (towards schemes) authorize refunds, and this would be confusing.

"DMS refunds"? We use "debit" as the terminology for "SMS" ... what's our terminology for "DMS"? We have none. That's the culprit of our naming problem, I think.

mt-clearhaus: I don't like "Authorization refunds" either, and I think it's a big step to introduce SMS and DMS :thinking: 

As a first stab I've pushed https://github.com/clearhaus/gateway-api-docs/pull/146/commits/ab4a3b0edccd603e9acd3560d4eef078f4f693e9.

website/content/gateway/api_reference/resources/debit_refunds/debit_refunds.md
mt-clearhaus: I've applied your suggestion regarding settlement vs clearing. It seems more correct than the original text, and we can change our minds later.

I will move on as we can introduce decline of late debit refunds later.

website/content/gateway/api_reference/resources/resources.md
mt-clearhaus: I'll allow myself to resolve this thread. We can pick it up later in another PR.

mt-clearhaus:
Rendering as of https://github.com/clearhaus/gateway-api-docs/pull/146/commits/594ada957155440f3923eec7f86c7ba579156e9f:

Screenshot from 2023-01-16 15-23-14
Screenshot from 2023-01-16 15-23-34
Screenshot from 2023-01-16 15-23-55
Screenshot from 2023-01-16 15-24-06
Screenshot from 2023-01-16 15-24-22

cbobach:
A couple of small things:
- `sender[country]` is stated as 2-3 chars in the [Visa API](https://developer.visa.com/capabilities/visa_direct/reference#tag/Funds-Transfer-API/operation/createPullFundsTransaction_v1%20-%20Latest!ct=[object%20Object]&path=senderCountryCode&t=request).
- `sender[postal_code]` is stated just as a string in the [Visa API](https://developer.visa.com/capabilities/visa_direct/reference#tag/Funds-Transfer-API/operation/createPullFundsTransaction_v1%20-%20Latest!ct=[object%20Object]&path=senderPostalCode&t=request).

Otherwise 👍🏻

website/content/gateway/api_reference/resources/debit_refunds/debit_refunds.md
```diff
@@ -0,0 +1,19 @@
...
+
+```shell
+POST https://gateway.clearhaus.com/debits/:id/refunds
+```
+
+##### Parameters
+No parameters are needed to make a new refund transaction.
+
+{{% notice %}}
+**Notice**: A refund of a Visa AFT debit might be declined by Visa or the issuer if the debit age is 24 hours or more, or the debit has cleared. The transaction gateway does not track the clearing of the debit, i.e. it relies on Visa and the issuer to decline late refunds.
```
tjconcept, line 18: It's not clear to me what the consequence of 

> The transaction gateway does not track the clearing of the debit

is.

Would this mean that an approved transaction isn't necessarily successfully cleared, and it would be on me to later reconcile with the merchant API whether the funds are deducted from my account or not?

mt-clearhaus: When an issuer, or Visa on behalf of an issuer via STIP (stand in), approves a debit (AFT), we expect the debit to be cleared eventually. Because this is SMS, not DMS, we don't send a separate clearing "request" upstream for a debit.
As with DMS, it would probably be incorrect/naive to expect that nothing can go wrong with the clearing, but we strongly expect it to go through.
When a debit is eventually cleared, the merchant API will know about it. The transaction gateway will not.
Visa states that a debit refund (AFTR) must be used to refund within the first 24 hours of the debit but before the clearing of the transaction. After this vague limit another mechanism ought to be used (we don't plan to support that at this time).
Since the transaction gateway doesn't know about clearing, at least currently, we have decided for now to let upstream (Visa/issuer) decide whether a "late" debit refund is approved or not.
I hope this makes matters a bit more clear.

mt-clearhaus: Could this be an improvement?

```suggestion
**Notice**: In almost all cases, an approved debit is eventually cleared by Visa and the issuer. It should be noted that Visa or the issuer might decline a debit refund with the reasoning that 24 hours have passed since the debit or that the debit has been cleared. There is no dedicated signaling for this, so the transaction has no dedicated status code. We recommend that a debit refund is created as soon as possible, in case a debit needs to be refunded.
```

cc: @ct-clearhaus @clearhaus/transaction-compliance

mt-clearhaus:
@cbobach [wrote](https://github.com/clearhaus/gateway-api-docs/pull/146#issuecomment-1384465908):

> `sender[country]` is stated as 2-3 chars in the [Visa API](https://developer.visa.com/capabilities/visa_direct/reference#tag/Funds-Transfer-API/operation/createPullFundsTransaction_v1%20-%20Latest!ct=[object%20Object]&path=senderCountryCode&t=request).

I chose to go with [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) as that's what we use elsewhere in the API (see https://docs.gateway.clearhaus.com/#account). We'll do the conversion to numeric before sending the country code upstream.

> `sender[postal_code]` is stated just as a string in the [Visa API](https://developer.visa.com/capabilities/visa_direct/reference#tag/Funds-Transfer-API/operation/createPullFundsTransaction_v1%20-%20Latest!ct=[object%20Object]&path=senderPostalCode&t=request).

This is a case where Visa's documentation is inconsistent. The item you point to in the POST Create Pull Funds Transaction resource:

Screenshot from 2023-01-17 10-19-33

However, in the AFT Processing Guide:

Screenshot from 2023-01-17 10-20-32
Screenshot from 2023-01-17 10-20-49

It seems they "got it right" in the [POST Create Push Funds Transaction (credit/OCT) resource](https://developer.visa.com/capabilities/visa_direct/reference#tag/Funds-Transfer-API/operation/createPushFundsTransaction_v1%20-%20Latest!ct=[object%20Object]&path=senderPostalCode&t=request):

Screenshot from 2023-01-17 10-23-44

So I think we should go with the `{5,10}` length requirement :man_shrugging:

cbobach (approved):

mt-clearhaus:
Rendering as of https://github.com/clearhaus/gateway-api-docs/pull/146/commits/7b1d516e710cee88470a24936a5557f51b58389f:

Screenshot from 2023-01-18 11-29-48
Screenshot from 2023-01-18 11-30-20
Screenshot from 2023-01-18 11-30-51
Screenshot from 2023-01-18 11-31-05
Screenshot from 2023-01-18 11-31-16

cbobach (approved):

website/content/gateway/api_reference/resources/debit_refunds/debit_refunds.md
tjconcept: > with DMS, it would probably be incorrect/naive to expect that nothing can go wrong with the clearing, but we strongly expect it to go through

Maybe I'm just in error, but I, as an integrator, would assume that liability to be on Clearhaus, once the gateway responds "OK" to the capture. Basically that's the guarantee extended from the authorization: the money is safe within  the ruleset of the scheme and your acquirer agreement. That's why merchants dare ship the goods on capture or otherwise debit a customer.
If the clearing fails, I'd expect that to be opaque to the merchant. E.g. not something we'd need to backfill from the merchant API and notify the merchant somehow that they won't get their funds anyway.

So the question is, I guess, is that generally wrong, or if not, is it likewise with AFTs, or would merchants assume some new type of risk that we'd need to have a technical answer to?

On another, more curiosity note, do you have a guess why "might be declined by Visa or the issuer if ... the debit has cleared" is the case? That makes their use quite limited, right? And very different from "capture refunds"?

mt-clearhaus: @tjconcept You bring up an interesting issue and some good questions. Sincere thank you for doing so.

I am not an expert on these matters, so I have discussed it with a few colleagues.

Rejects connected to clearing can also happen for DMS, i.e. for our current captures and refunds, and we do not reflect that in the transaction gateway API either. In other words, debits/AFTs do not pose a new challenge here. Handling these rejects is a partly manual process, where the involved parties' state is reverted. Fortunately, we do not see many rejections.

> would merchants assume some new type of risk that we'd need to have a technical answer to?

With regard to risk, we do not assume debits/AFTs to be different from the current authorizations, captures and refunds (DMS). However, since we have no history and experience some difficulty in getting firm statements from Visa, we cannot confirm it.

> do you have a guess why "might be declined by Visa or the issuer if ... the debit has cleared" is the case? That makes their use quite limited, right? And very different from "capture refunds"?

We are not entirely sure, however, as for AFT use cases it seems natural that you deliver quickly and, if that fails, you also refund quickly. We imagine within seconds or at most a few minutes.

An option to refund later exists in Visa's machinery, however, this option is another transaction type that is apparently badly supported among issuers as many or all them process the transactions manually with quite a delay (probably several days or even weeks). We do not support this option.

tjconcept: Okay, thanks. I'll bring it up with our partner contact when discussing our beta merchant's use case. In short, we either need to invest significantly in as-recent-as-possible reconciliation to build our own "notification" of clearing issues (as we could be losing uncapped volumes) while we test and get a sense of the risk, or Clearhaus needs to cover it one way or another.
I'd argue any day for moving that risk upstream as far as possible (making all downstreams simpler), but that's unfortunately "business" and not just rational, I suppose.

> is another transaction type

Oh well 😅

mt-clearhaus:
Rendering as of https://github.com/clearhaus/gateway-api-docs/pull/146/commits/98d34329ac64633ba1499adf0c3547836f68cc2a:

Screenshot from 2023-01-23 21-22-11
Screenshot from 2023-01-23 21-22-51
Screenshot from 2023-01-23 21-23-14
Screenshot from 2023-01-23 21-23-31
Screenshot from 2023-01-23 21-23-44
Screenshot from 2023-01-23 21-23-58

mt-clearhaus:
Added in https://github.com/clearhaus/gateway-api-docs/pull/146/commits/d1bab520261bcb3cb1b0bcb002efcf4754b36951:

Screenshot from 2023-01-31 15-03-32

The links work :heavy_check_mark:

mt-clearhaus:
Rendering as of https://github.com/clearhaus/gateway-api-docs/pull/146/commits/33f63b6d300901a8028df57da8ef6ebcc1d04d75:

Screenshot from 2023-02-01 16-31-17
Screenshot from 2023-02-01 16-31-49
Screenshot from 2023-02-01 16-32-52
Screenshot from 2023-02-01 16-33-06
Screenshot from 2023-02-01 16-33-21
Screenshot from 2023-02-01 16-33-39

cbobach (approved):

PR #148, opened 2022-06-20
Make the example signing API key less realistic

ct-clearhaus:
Notice that both version and variant are unchanged:

```
0a1b2c3d-4e5f-4c2f-9999-000000000000
   version ---^    ^--- variant
```

* Version: 4, because hex char 13 is still a 4.
* Variant: 1, because hex char 17 is still 8-b.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,19 +3,13 @@ version: 2
 jobs:
   build:
     machine:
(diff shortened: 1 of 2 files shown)


mt-clearhaus (approved):
I see no harm in this :slightly_smiling_face: :+1:

PR #150, opened 2022-07-19
Explain default behaviour for test pans

ct-clearhaus:
... when the transaction amount (which is used to trigger specific status codes) is not a valid status code.


ct-clearhaus:
Originated from [slack](https://clearhaus.slack.com/archives/C02AEM3596X/p1658221047301779). Feel free to merge without further ado 🙂

brorbw (approved):

PR #157, opened 2022-10-26
Make token one payment method

mt-clearhaus:
We do not consider `token[m4m]` and `token[vts]` to be two different payment methods. M4M and VTS are token frameworks living under the `token` payment method.

Feel free to merge.


ct-clearhaus (approved):
👍

PR #158, opened 2022-11-08
Add optional token[m4m][rcai] parameter

mt-clearhaus:
Remote Commerce Acceptor Identifier. At present only relevant for Mastercard, i.e. only for `token[m4m]`, not for `token[vts]`.

Rendered:

Screenshot from 2022-11-08 20-58-36

The two examples for the identifier are borrowed from Mastercard's Customer Interface Specification (DE104SE003).

I found the parameter for SCOF here:
https://developer.mastercard.com/secure-card-on-file/documentation/new-api-reference/apis/#checkout

I have established that the parameter is not present for MDES here:
https://developer.mastercard.com/mdes-digital-enablement/documentation/api-reference/

Mastercard writes that the parameter is "expected to be present in remote commerce transactions initiated via wallets or remote commerce programs that support DSRP cryptogram format 3". I have simplified this to "optional" as we do not dive into such processing details in our API documentation.

### DIFF
--- a/website/content/gateway/api_reference/resources/authorizations/authorizations_methods/token.md
+++ b/website/content/gateway/api_reference/resources/authorizations/authorizations_methods/token.md
@@ -56,6 +56,15 @@ Found in `encryptedPayload.encryptedData.de48se43Data` (MDES) or `encryptedPaylo
 {{% regex_optional %}}Required for CITs; otherwise optional.{{% /regex_optional %}}
 {{% /description_details %}}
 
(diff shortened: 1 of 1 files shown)


AlexanderIversen:
LGTM :)

ad-clearhaus (approved):
I am a bit out of the loop regarding _second phase_, so I don't have the context behind the changes. But the additions looks correct to me. :+1:

ab-clearhaus (approved):
It looks correct and is hereby approved from my side.

mt-clearhaus:
This has been deployed in the production environment :heavy_check_mark:

PR #163, opened 2022-12-21
Remove HRK from currency list

ct-clearhaus:
Should not be merged before 2023-02-15.


thejspr:
HRK is already removed

PR #165, opened 2023-01-10
Remove babel/core

Lassejoe:
Due to: https://github.com/clearhaus/gateway-api-docs/security/dependabot?q=is%3Aopen+manifest%3Ayarn.lock+package%3Ajson5

We don't use babel/core for anything so we can just remove it for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,6 @@
     "imagemin":"^7.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


mt-clearhaus (approved):
I've tested locally that building and serving still works :heavy_check_mark: 

I see this means `json5` disappears :heavy_check_mark: 

LGTM :+1:

PR #166, opened 2023-02-01
Retry response message proposal

brorbw:
A first stab at a retry message to indicate if a transaction can be retried or not. Feel free to scrutinized.


brorbw:
Alternative: https://github.com/clearhaus/gateway-api-docs/pull/167

brorbw:
Third alternative: https://github.com/clearhaus/gateway-api-docs/pull/168

PR #46, opened 2017-09-27
Add CLP/UGX exponent change

ct-clearhaus:

PR #170, opened 2023-04-12
Add ISK exponent change to Changes section

ct-clearhaus:
As communicated in the 2023Q1 Partner Bulletin.

### DIFF
--- a/website/content/gateway/api_reference/changes/changes.md
+++ b/website/content/gateway/api_reference/changes/changes.md
@@ -10,7 +10,7 @@ Follow coming changes on the [source code repository](https://github.com/clearha
 
 Sorted by descending timestamp.
 
-#### ISK exponent change
+#### Icelandic Krona (ISK) exponent change
 ISK changes exponent from 2 to exponent 0.
 Transactions in ISK will be declined between 2023-04-13T19:00:00Z (Thursday
 evening) and 2023-04-17T08:00:00Z (Monday morning) (both inclusive); before this


ct-clearhaus:
This one will conflict with https://github.com/clearhaus/gateway-api-docs/pull/169. I don't mind which one goes first and the sorting of the two entries 🤷‍♂️

ct-clearhaus:
Much like https://github.com/clearhaus/gateway-api-docs/pull/46 btw.

website/content/gateway/api_reference/changes/changes.md
```diff
@@ -10,6 +10,12 @@ Follow coming changes on the [source code repository](https://github.com/clearha
 
 Sorted by descending timestamp.
 
+#### ISK exponent change
```
ad-clearhaus, line 13:
```suggestion
#### Icelandic Krona (ISK) exponent change
```

ad-clearhaus: To reflect the format in

> Croatian Kuna (HRK) changes to Euro (EUR)

But then I notice we have other announcements

> Change SLL currency to SLE

> Removing VES from supported currencies

I guess we haven't been that stringent in the past.

I prefer writing out "Icelandic Krona (ISK)" but am happy with any version.

tjconcept:
Is this going to impact the merchant API?

I'm generally in doubt how to interpret amounts in the merchant API. I tried finding it in the documentation. I think we currently assume your APIs to use the same format, but that may be a mistake in the first place.

hovmand:
@tjconcept I assume you're mainly interested in Transactions and Disputes? i.e.: https://developer.clearhaus.com/rels/transaction#properties and https://developer.clearhaus.com/rels/dispute#properties

For both of these resources the `amount` is in minor units. There's a note attached to both of these properties:
> Amount in minor units of given currency (e.g. cents if in Euro).

Your assumption is correct. It must be interpreted the same way as the `amount` that is sent to the gateway when creating a transaction.


Looking at "amounts" for other resources throughout the API (e.g. an Application) I can see that it's not consistent. If you need an explanation for any other properties/resources please let me know.

tjconcept:
So, are all ISK amounts from the API, historical and future resources, now formatted without minor units? And if so, have you truncated the amount and thereby "discarded" the historic information about minor units?

tjconcept:
Okay, thanks.

To double check, if I understand you correctly, in order to interpret the data correctly, I should switch on the processing date of transactions or disputes (dispute or transaction date in this case?) to either consider it to contain two decimal digits or zero?

I'm also interested in settlement data 🙂

> There's a note attached to both of these properties

I'll humbly suggest you add a link to some explainer about this and possibly even the historical changes one must implement.

hovmand:
You should use the following properties for switching between exponents:
- Disputes: `opened_at`
- Transactions: `processed_at`

As we're not settling in ISK, Settlement amounts shouldn't be an issue?

---

> I'll humbly suggest you add a link to some explainer about this and possibly even the historical changes one must implement.

Duly noted.

PR #173, opened 2023-05-02
Make name on card mandatory for POW for mastercard

brorbw:
From the 6th of June payment of winnings for Mastercard will require name of recipient.

Internal link: https://github.com/clearhaus/issues-project-management/issues/102

### DIFF
--- a/website/content/gateway/examples/payout_to_cardholder/payout_to_cardholder.md
+++ b/website/content/gateway/examples/payout_to_cardholder/payout_to_cardholder.md
@@ -33,5 +33,5 @@ Example response (snippet):
 }
 ```
 
-The field {{% highlight_text %}}card[name]{{% /highlight_text %}} is mandatory for Mastercard payment of winnings (gaming/gambling merchants).
+The field {{% highlight_text %}}card[name]{{% /highlight_text %}} is mandatory for Mastercard Payment of winnings (gaming/gambling merchants).
 Depending on card scheme and merchant category, the name on the card might be necessary for approval of credits.
--- a/website/content/gateway/examples/payout_to_cardholder/payout_to_cardholder.md
+++ b/website/content/gateway/examples/payout_to_cardholder/payout_to_cardholder.md
@@ -32,4 +32,6 @@ Example response (snippet):
     "currency": "EUR"
 }
 ```
-Depending on card scheme and merchant category, the name on the card might be necessary for approval of credits. It may be provided through the optional parameter {{% highlight_text %}}card[name]{{% /highlight_text %}}.
+
+The field {{% highlight_text %}}card[name]{{% /highlight_text %}} is mandatory for Mastercard payment of winnings (gaming/gambling merchants).
+Depending on card scheme and merchant category, the name on the card might be necessary for approval of credits.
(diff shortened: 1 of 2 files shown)


website/content/gateway/examples/payout_to_cardholder/payout_to_cardholder.md
```diff
@@ -32,4 +32,6 @@ Example response (snippet):
     "currency": "EUR"
 }
 ```
-Depending on card scheme and merchant category, the name on the card might be necessary for approval of credits. It may be provided through the optional parameter {{% highlight_text %}}card[name]{{% /highlight_text %}}.
+
+The field {{% highlight_text %}}card[name]{{% /highlight_text %}} is mandatory for Mastercard payment of winnings (gaming/gambling merchants).
```
cbobach, line 36:
```suggestion
The field {{% highlight_text %}}card[name]{{% /highlight_text %}} is mandatory for Mastercard Payment of Winnings (gaming/gambling merchants).
```
To align with API reference edit.

cbobach (commented):
I agree with @mt-clearhaus that it was required in the API reference section. I think the addition you have made is really good.

Ready to approve if no-one else has something to add.

website/content/gateway/examples/payout_to_cardholder/payout_to_cardholder.md
brorbw:
```suggestion
The field {{% highlight_text %}}card[name]{{% /highlight_text %}} is mandatory for Mastercard Payment of winnings (gaming/gambling merchants).
```

brorbw:
Maybe I can get you @ct-clearhaus to take a gander?

ct-clearhaus:
Uhm, my instincts says: if we have a way to err like `card[name] is required for mastercard payment of winnings`, then I agree that the API reference should state that the parameter is not "Optional" but "Required for Mastercard Payment of winnings, otherwise optional". Otherwise, it's API-wise, technically speaking, optional — although approval rate may be very low (depending on your traffic).
Further "non-API technical" (but rather "compliance requirement") details could also be put in the partner guideline (and referenced from here).

WDYT?

brorbw:
Afaict credits without name on card will be rejected. I have no evidence for this other than the field being required. If that is the case it would be nice if we could err on those types of transactions. I'm unsure how early in the request we know if it's for gaming gambling but we might be able to put together a filter to err on those transactions.

mt-clearhaus:
I @ct-clearhaus is referring to the code in the transaction gateway. There is currently no code to decline/reject a credit if `card[name]` is not supplied. And he hints the API reference in our docs is usually about what the transaction gateway requires, not what upstream requires to approve.

brorbw:
Maybe I was unclear. I'm trying to say that is that we should build a filter such that we reject such transactions and that is what I'm hinting with this PR. It could also just be a note and make the parameter optional.
I don't think we should accept gaming gambling payouts for mastercard without name on card.

brorbw:
https://github.com/clearhaus/klovedal/pull/725

mt-clearhaus (approved):
:+1:

PR #174, opened 2023-05-04
Add refunds, debits and debit refunds to endpoint summary

cbobach:
I've found that we are missing some endpoint in Endpoint summary.
image

This will add them. Served locally:
image


mt-clearhaus (approved):
Nice catch :+1:

PR #175, opened 2023-05-31
Explicitly state that amount must be greater than zero.

cbobach:
Visa Direct does not allow for debits with amount zero. This will be stated by this update to our documentation.

I've remove the exception related to 3DSv1 as this was removed from the documentation elsewhere by https://github.com/clearhaus/gateway-api-docs/commit/a93ddde0210227935ccf129f192a29316f3c459f


cbobach:
I haven't added a change summary under changes as there is no change in the way the API behaves but solely a precision in the documentation. Please comment if you think I'm wrong with regards to this.

mt-clearhaus (approved):
:+1: 

I agree this does not warrant a changelog entry.

PR #179, opened 2023-09-22
Extend COF parameter explanation

ct-clearhaus:
Only the "store" value was explained in text; the added text is addressing the "use" value.


mt-clearhaus (approved):
:+1:

PR #181, opened 2023-10-26
Update googlepay[recipient_id] description

cbobach:
Related issue https://github.com/clearhaus/issues-pci/issues/4156#event-10765079084

New rendered version
image

As is
image

The note still state how to use it in test
image

I think this hints towards how to use it.


mt-clearhaus (approved):
:+1:

PR #182, opened 2023-10-31
Mention funding transactions in debits notice block

mt-clearhaus:
The original reason for this was that debits are not supported for marketplace merchants, however, the more "general" reasoning is that our debits currently support funding transactions (on Visa cards).


cbobach (approved):

AlexanderIversen (commented):
Maybe we should consider using a phrase like "Debits are only supported for Visa Cards and for Funding Transactions." 
The reason is that it is not only marketplace transactions that we do not support with debits but all Purchase transactions meaning all transactions where a product or service is bought. 

If so I might need to update the Partner Guideline with some clarification on Purchase, Funding and Quasi-cash transactions.

mt-clearhaus:
@AlexanderIversen I had to read your reasoning twice but I think I get it now :+1: :bow: And I agree. Just like we assume the reader knows what a marketplace is, in the API documentation, I suppose can also assume the reader knows what a Funding Transaction is. I'll use your suggested sentence.

AlexanderIversen (approved):
LGTM. :)

PR #183, opened 2023-12-15
Add sca_exemption[transaction_risk_analysis]

mt-clearhaus:
To support the indication of a Transaction Risk Analysis (TRA) SCA exemption upstream.

Created as a dictionary because we will eventually support other exemption types. In addition, we don't want to close the door on the possibility for signaling more than one exemption type on a particular transaction, even though it seems card schemes currently advise against it.

### DIFF
--- a/website/content/gateway/api_reference/resources/authorizations/authorizations.md
+++ b/website/content/gateway/api_reference/resources/authorizations/authorizations.md
@@ -70,8 +70,7 @@ Default:
 - `recurring`\: A series of transactions where the cardholder has explicitly agreed that the merchant may repeatedly charge the cardholder at regular, predetermined intervals that may not exceed 1 year.
 - `unscheduled`\: A series of transactions where the cardholder has explicitly agreed that the merchant may repeatedly charge the cardholder at non-predetermined times, e.g. based on cardholder usage.
 
-{{% regex_optional %}}Conditional.{{% /regex_optional %}}
-Cannot be present if `series[previous]` is present.
+{{% regex_optional %}}Mutually exclusive with `series[previous]` and `sca_exemption`.{{% /regex_optional %}}
 
 {{% /description_details %}}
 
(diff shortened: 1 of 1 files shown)


mt-clearhaus:
Not all exemption types we might support in the future abbreviate well so I have decided to not abbreviate Transaction Risk Analysis after all. Input is very welcome.

Current rendering:

Screenshot from 2023-12-15 11-47-15

website/content/gateway/api_reference/resources/authorizations/authorizations.md
```diff
@@ -85,6 +85,10 @@ If the latest approved authorization in the series was not processed via Clearha
 Conditional. Cannot be present if `series[type]` is present.
 {{% /description_details %}}
 
+{{% description_term %}}sca_exemption[transaction_risk_analysis] {{% regex %}}(true|false){{% /regex %}}{{% /description_term %}}
+{{% description_details %}}Whether or not to claim a Transaction Risk Analysis (TRA) SCA exemption.
```
mt-clearhaus, line 89:
```suggestion
{{% description_details %}}Whether or not to claim Transaction Risk Analysis (TRA) SCA exemption.
```

:man_shrugging:

mt-clearhaus: Alternative:

```suggestion
{{% description_details %}}Whether or not to request Transaction Risk Analysis (TRA) SCA exemption.
```

mt-clearhaus:
Current rendering:

Screenshot from 2023-12-18 10-27-46

AlexanderIversen (commented):
LGTM. The only thought I had was whether or not we should mention that TRA is something you should be allowed to utilize as a partner. 

We could also argue that it is something that should be writing in the Partner Guideline.

website/content/gateway/api_reference/resources/authorizations/authorizations.md
```diff
@@ -85,6 +85,10 @@ If the latest approved authorization in the series was not processed via Clearha
 Conditional. Cannot be present if `series[type]` is present.
 {{% /description_details %}}
 
+{{% description_term %}}sca_exemption[transaction_risk_analysis] {{% regex %}}(true|false){{% /regex %}}{{% /description_term %}}
+{{% description_details %}}Whether or not to request Transaction Risk Analysis (TRA) SCA exemption.
+{{% /description_details %}}
+
```
cbobach, line 91: Is this the right place to place it? If we won't end up supporting many different sca exemption types I think so, but if we already know that we will end up with 3 or more I think we should create a section dedicated to sca exemptions.

mt-clearhaus: I lean towards making such "refactoring" when it is needed and not now.

I chose this place because the parameters appear to be in alphabetic order.

cbobach (dismissed):
:muscle:

mt-clearhaus:
@AlexanderIversen I tend to point to the Partner Guideline. In this documentation we currently don't have many places with such notice. We have a single place with "Optional when partner is trusted." (`card[csc]`), however, in this case it will not be `trusted` (true/false) that decides on the feature's availability.

mt-clearhaus:
Current rendering:

Screenshot from 2023-12-22 13-32-04

ct-clearhaus:
> @AlexanderIversen: The only thought I had was whether or not we should mention that TRA is something you should be allowed to utilize as a partner.

> @mt-clearhaus: (...) point to the Partner Guideline. In this documentation we currently don't have many places with such notice.

We have "Please be aware that rules to disallow captures for such authorizations may be in place for a merchant." in a "Notice" box. I think it would be fine to have something similar in this case.

mt-clearhaus:
Current rendering:

Screenshot from 2023-12-22 13-43-41

mt-clearhaus:
We will continue with https://github.com/clearhaus/gateway-api-docs/pull/184, i.e. this PR can be closed.

PR #186, opened 2024-02-20
Add notice to sca_exemption parameter

mt-clearhaus:
To make it clear we honor the request even in a situation where SCA is not required, e.g. one leg out.

Rendering:

Screenshot from 2024-02-20 09-09-29


AlexanderIversen (approved):
I agree that it makes sense to mention this and I would also say this is the right decision. Not because the fraud limit for TRA should be calculated for merchants/PSP outside EEA but because we should expect the EMV 3DS protocol to be more widely supported outside the EEA, and therefore I think it's more future proof to accept TRA exemption requests even thought it is an OLO transaction.

PR #198, opened 2024-10-01
Document voiding of failed authorizations

thejspr:
Issue: https://github.com/clearhaus/issues-pci/issues/5094


website/content/gateway/api_reference/changes/changes.md
```diff
@@ -10,6 +10,9 @@ Follow coming changes on the [source code repository](https://github.com/clearha
 
 Sorted by descending timestamp.
 
+#### Allowing voids on status code 50000 authorizations
+We've added support for voids on authorizations with status code 50000.
```
mt-clearhaus, line 14:
```suggestion
We have added support for voids on authorizations with status code 50000.
```

Just to be a bit more formal. I think this is the current style of the documentation (and IMHO it should be).

cbobach (approved):

PR #201, opened 2024-12-19
Clarify applepay[raw][eci] is required for Visa

mt-clearhaus:
This is currently the case and should be reflected by the API documentation.

Rendering:

Screenshot from 2024-12-19 11-03-10


cbobach (approved):

mt-clearhaus:
I think it makes sense to merge this because this is how the transaction gateway currently works.

PR #204, opened 2025-04-09
Compliance updates visa 2.6

brorbw:
Changes to currency code for Curaçao and Sint Maarten


website/content/currencies.md
```diff
@@ -7,12 +7,13 @@ anchor: "currencies"
 weight: 230
 ---
 | Code | Num | Exp | Comment                      |
-| ---- | --- | --- | ---------------------------- |
+|------|-----|-----|------------------------------|
```
mt-clearhaus, line 10: Is this really correct? :thinking: 

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-tables

brorbw: No. It's not. I mean it still renders, but my `format-on-save-hook` got the best of me.

website/content/currencies.md
```diff
@@ -12,7 +12,8 @@ weight: 230
 | AFN  | 971 | 2   |                              |
 | ALL  | 008 | 2   |                              |
 | AMD  | 051 | 2   |                              |
-| ANG  | 532 | 2   |                              |
+| ANG  | 532 | 2   | Not recognized by Visa       |
+| XCG  | 532 | 2   | Not recognized by Mastercard |
```
mt-clearhaus, line 16: Wouldn't it make sense to keep the table sorted on the alpha code?

brorbw: I wanted it to be easy to see that they belong together so I purposely ignored the sorting for this entry. I think it makes a bit more sense to keep them together. This is the only duplicate code so there is no precedence.

mt-clearhaus (approved):

PR #240, opened 2026-06-02
Add payload and assurance data parameters for click to pay

MikkelThingholm:
Suggestion for simplifying click to pay and 3D secure flow. We add two new parameters `payload` and `assurance_data` for the click to pay payment method:
image

### DIFF
--- a/website/assets/css/docs.css
+++ b/website/assets/css/docs.css
@@ -940,6 +940,19 @@ div.docs.scrollpage section.content .wrapper {
 	padding: 0 30px;
 }
 
+div.docs .deprecated-badge {
+	align-self: flex-start;
+	margin-left: 10px;
+	padding: 2px 6px;
+	border-radius: 4px;
+	color: var(--color-brand-orange-60);
+	background-color: var(--color-brand-orange-10);
+	font-size: 13px;
+	font-weight: 600;
+	text-transform: uppercase;
+	transform: translateY(4px);
+}
+
 .mobile-navigation-content {
 	display: none;
 }
--- a/website/content/gateway/api_reference/changes/changes.md
+++ b/website/content/gateway/api_reference/changes/changes.md
@@ -10,6 +10,12 @@ Follow coming changes on the [source code repository](https://github.com/clearha
 
 Sorted by descending timestamp.
 
+#### Updates to Click to Pay
+Starting 2026-xx-xx, we will expand the Click to Pay payment method:
+- The payment method allows for both cards and tokens.
+- Transactions can be created using a payload interface in addition to the raw interface.
+- The raw interface supports `[onbehalf][...]` fields to be used when 3-D Secure authentication was performed `ONBEHALF`.
+
 #### Additional SCA exemtions added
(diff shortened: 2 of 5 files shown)


nikiepaydk:
I dont know if you take PSP feedback on this. But I would really like for this change to not happen.

It is much simpler for us as PSP's to integrate with acquirers when they keep to the normal seperate pan/tan, expiry, cryptogram and 3ds fields.

We already have a lot of extra code, that is only necessary for supporting Clearhaus among all of our supported acquirers due to these "simplifying api's" where you expect the raw payloads for things such as googlepay, applepay and 3DS. Where other acquireres just request us to provide the normal extracted common fields.

Your current ClickToPay interface is MUCH simpler for us to integrate. This change will increase our integration burden.

This change also assumes all click to pay implementations provide the full raw payloads, which I would argue cannot be assumed for all click to pay integrations.

cbobach:
@nikiepaydk thank you for your feedback, it is very much welcomed :pray: We'll take that into consideration going forward.

This is not to replace the already implemented `clicktopay` solution, but to add it as an alternative as we have other PSPs asking for this. You'll see that we for `applepay` also support both a 'direct' and 'raw' approach.

nikiepaydk:
@cbobach Okay great to hear. My fear was that this would be similar to MobilePay and GooglePay where you only support the full payload.

Just to make sure, the new descriptions seems to hint at a solution similar to MobilePay, where token authentication requires the full payload. Similar to these, the description seems to me to indicate the same, that if passkeys / Click To Pay managed 3DS is used, that the full payload is required.

Other acquirers have support for click to pay, by us simply providing the dsTransID and authenticationValue as seperate fields.

The solution with ApplePay where you provide both options is great and have made our ApplePay integration much simpler due to this :smile:

cbobach:
You input as a PSP is always very much appreciated :bow: 

As part of this implementation of Click to Pay we are also looking at providing a 'raw' input of 3DS fields, as we know that might align better.

nikiepaydk:
> As part of this implementation of Click to Pay we are also looking at providing a 'raw' input of 3DS fields, as we know that might align better.

That would be amazing!

MikkelThingholm:
@nikiepaydk Feedback is very much appreciated.

I have done a few updates to the proposed changes. Here is the full proposed click to pay documentation.
screencapture-localhost-1313-2026-06-04-09_00_47

nikiepaydk:
> @nikiepaydk Feedback is very much appreciated.
> 
> I have done a few updates to the proposed changes. Here is the full proposed click to pay documentation. ...

I could not asked for more! Thank you!

cbobach:
@nikiepaydk related to the raw 3DS values https://github.com/clearhaus/gateway-api-docs/pull/241

peterklogborg (dismissed):

MikkelThingholm:
Current proposition. I have added a `[onbehalf][...]` fields for 3-D Secure raw values when performed `ONBEHALF` and added support for cards in the documentation.
image
image

PR #243, opened 2026-06-18
Remove perl from docker image

peterklogborg:
I've removed perl since there was a lot of new vulnerabilities found in it. We don't need it, and the service runs without it.


peterklogborg:
not in pci scope and its vulnerabilities arent and wont be relevant

